PR #7, opened 2022-03-21
fix(Patcher+Io): fix reading input multiple times

Bleuzen:
- create copy of input as ByteArray
 - create new input stream for every read because they are not reusable

this partly fixes https://github.com/ReVancedTeam/revanced-patcher/issues/5


Lucaskyy (approved):

Lucaskyy:
This does read the entire input JAR into memory.
Is there no way to reset a JarInputStream/ZipInputStream?

Bleuzen:
will redo this tomorrow

Issue #5, opened 2022-03-20
Saving jar does not work


Bleuzen:

Bleuzen:
reopen because there is more to it, still needs additional fixes

PR #8, opened 2022-03-21
fix(Io): JAR loading and saving

oSumAtrIX:
## This PR addresses issue #5 and continues PR #7.

Io will now be instantiated with both, input as well as an output stream alongside a mutable class list:
https://github.com/ReVancedTeam/revanced-patcher/blob/0f3db838a05fb376fe974b4a1bb3292888f8c5a0/src/main/kotlin/net/revanced/patcher/util/Io.kt#L13-L17

The streams are used until we call `saveAsJar`. The original streams will be closed:
https://github.com/ReVancedTeam/revanced-patcher/blob/0f3db838a05fb376fe974b4a1bb3292888f8c5a0/src/main/kotlin/net/revanced/patcher/util/Io.kt#L86-L90

When writing, the input stream will be proxied through a  `BufferedInputStream` a second time, all classes will be merged with all non-class file zip entries and written into the output stream:
https://github.com/ReVancedTeam/revanced-patcher/blob/0f3db838a05fb376fe974b4a1bb3292888f8c5a0/src/main/kotlin/net/revanced/patcher/util/Io.kt#L56-L84

### All tests and usages in the patcher have been adjusted accordingly.

### DIFF
--- a/src/main/kotlin/net/revanced/patcher/util/Io.kt
+++ b/src/main/kotlin/net/revanced/patcher/util/Io.kt
@@ -26,10 +26,10 @@ internal class Io(
 
         // read all entries from the input stream
         // we use JarEntry because we only read .class files
-        var jarEntry: JarEntry?
-        while (jis.nextJarEntry.also { jarEntry = it } != null) {
+        lateinit var jarEntry: JarEntry
+        while (jis.nextJarEntry.also { if (it != null) jarEntry = it } != null) {
             // if the current entry ends with .class (indicating a java class file), add it to our list of classes to return
-            if (jarEntry!!.name.endsWith(".class")) {
+            if (jarEntry.name.endsWith(".class")) {
                 // create a new ClassNode
                 val classNode = ClassNode()
                 // read the bytes with a ClassReader into the ClassNode
@@ -59,7 +59,7 @@ internal class Io(
             if (zipEntry.name.endsWith(".class")) continue
 
             // create a new zipEntry and write the contents of the zipEntry to the output stream
-            jos.putNextEntry(ZipEntry(zipEntry.name))
+            jos.putNextEntry(ZipEntry(zipEntry))
             jos.write(jis.readBytes())
 
             // close the newly created zipEntry
--- a/src/main/kotlin/net/revanced/patcher/util/Io.kt
+++ b/src/main/kotlin/net/revanced/patcher/util/Io.kt
@@ -16,13 +16,12 @@ internal class Io(
     private val input: InputStream,
     private val output: OutputStream,
     private val classes: MutableList<ClassNode>
-    )
-{
+) {
     private val bufferedInputStream = BufferedInputStream(input)
 
     fun readFromJar() {
         bufferedInputStream.mark(0)
-       // create a BufferedInputStream in order to read the input stream again when calling saveAsJar(..)
+        // create a BufferedInputStream in order to read the input stream again when calling saveAsJar(..)
         val jis = JarInputStream(bufferedInputStream)
 
         // read all entries from the input stream
--- a/src/main/kotlin/net/revanced/patcher/util/Io.kt
+++ b/src/main/kotlin/net/revanced/patcher/util/Io.kt
@@ -55,7 +55,7 @@ internal class Io(
         var zipEntry: ZipEntry?
         while (jis.nextEntry.also { zipEntry = it } != null) {
             // skip all class files because we added them in the loop above
-            // TODO: check for zipEntry.isDirectory
+            // TODO(oSumAtrIX): Check for zipEntry.isDirectory
             if (zipEntry!!.name.endsWith(".class")) continue
 
             // create a new zipEntry and write the contents of the zipEntry to the output stream
--- a/src/main/kotlin/net/revanced/patcher/util/Io.kt
+++ b/src/main/kotlin/net/revanced/patcher/util/Io.kt
@@ -52,6 +52,7 @@ internal class Io(
 
         // first write all non .class zip entries from the original input stream to the output stream
         // we read it first to close the input stream as fast as possible
+        // TODO(oSumAtrIX): There is currently no way to remove non .class files.
         lateinit var zipEntry: ZipEntry
         while (jis.nextEntry.also { if (it != null) zipEntry = it } != null) {
             // skip all class files because we added them in the loop above
--- a/src/main/kotlin/net/revanced/patcher/util/Io.kt
+++ b/src/main/kotlin/net/revanced/patcher/util/Io.kt
@@ -3,47 +3,92 @@ package net.revanced.patcher.util
 import org.objectweb.asm.ClassReader
 import org.objectweb.asm.ClassWriter
 import org.objectweb.asm.tree.ClassNode
+import java.io.BufferedInputStream
 import java.io.InputStream
 import java.io.OutputStream
 import java.util.jar.JarEntry
 import java.util.jar.JarInputStream
-import java.util.jar.JarOutputStream
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+import java.util.zip.ZipOutputStream
 
-object Io {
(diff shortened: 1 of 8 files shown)


src/test/kotlin/net/revanced/patcher/ReaderTest.kt
```diff
@@ -1,12 +1,13 @@
...
+import com.sun.xml.internal.messaging.saaj.util.ByteOutputStream
 import kotlin.test.Test
 
 internal class ReaderTest {
     @Test
     fun `read jar containing multiple classes`() {
-        val testData = PatcherTest::class.java.getResourceAsStream("/test2.jar")!!
-        Patcher(testData, PatcherTest.testSigs) // reusing test sigs from PatcherTest
-        testData.close()
+        val testData = PatcherTest::class.java.getResourceAsStream("/test1.jar")!!
```
Lucaskyy, line 9:
```suggestion
        val testData = PatcherTest::class.java.getResourceAsStream("/test2.jar")!!
```
`test1.jar` only has 1 class. This test is to test if the implementation of `Io` can read multiple classes from a JAR file.
`test2.jar` actually has 2 classes, so this one needs to be used for this test.

src/test/kotlin/net/revanced/patcher/ReaderTest.kt
```diff
@@ -1,12 +1,13 @@
...
+        val outFile = ByteOutputStream()
+        Patcher(testData, outFile, PatcherTest.testSignatures) // reusing test sigs from PatcherTest
```
Lucaskyy, line 11: There's no need for the `outFile` parameter. You can inline it.

src/test/kotlin/net/revanced/patcher/ReaderTest.kt
```diff
@@ -1,12 +1,13 @@
 package net.revanced.patcher
 
+import com.sun.xml.internal.messaging.saaj.util.ByteOutputStream
```
Lucaskyy, line 3: Wrong import. Don't ever use classes from `com.sun.*`, it's recommended against by Oracle since these classes are implementation-bound. (this class only exists in certain JVMs, definitely not in Android at least)
```suggestion
import java.io.ByteArrayOutputStream
```

src/test/kotlin/net/revanced/patcher/PatcherTest.kt
```diff
@@ -110,39 +114,25 @@ internal class PatcherTest {
...
-    //    assertEquals(available, out.size())
-    //
-    //    out.close()
-    //    testData.close()
-    //}
+    @Test
+    fun `test patcher with no changes`() {
+        val patcher = Patcher(
+            PatcherTest::class.java.getResourceAsStream("/test1.jar")!!,
+            ByteArrayOutputStream(),
```
Lucaskyy, line 132: Make the output stream a variable and test that with no changes the output should be the same as the input. (you removed this code, see the PR diff to check how I did it previously or do it your own way)

src/test/kotlin/net/revanced/patcher/PatcherTest.kt
```diff
@@ -44,8 +45,11 @@ internal class PatcherTest {
 
     @Test
     fun testPatcher() {
-        val testData = PatcherTest::class.java.getResourceAsStream("/test1.jar")!!
-        val patcher = Patcher(testData, testSigs)
+        val patcher = Patcher(
+            PatcherTest::class.java.getResourceAsStream("/test1.jar")!!,
+            ByteArrayOutputStream(),
```
Lucaskyy, line 50: Tests need to be added for this output. You don't have to make ***this test*** now, it can be added later. (the test for no changes needs to be added, though)

src/main/kotlin/net/revanced/patcher/util/Io.kt
```diff
@@ -3,47 +3,90 @@ package net.revanced.patcher.util
...
+            jos.write(cw.toByteArray())
+
+            // close the newly created jar entry
             jos.closeEntry()
         }
+
+        // finally, close all streams
+        // TODO: order & necessity to close all of the streams check
+        jos.close()
+        bufferedInputStream.close()
```
Lucaskyy, line 88: Move this to under the `jis.close()` (above the `patchedClass` for-loop)

src/main/kotlin/net/revanced/patcher/util/Io.kt
```diff
@@ -3,47 +3,90 @@ package net.revanced.patcher.util
...
+        input.close()
```
Lucaskyy, line 89: This one too. Place it under the `bufferedInputStream.close()` above the for-loop again.

src/main/kotlin/net/revanced/patcher/util/Io.kt
```diff
@@ -3,47 +3,90 @@ package net.revanced.patcher.util
...
-                jos.write(jis.readBytes())
-            }
+    fun saveAsJar() {
+        val jis = ZipInputStream(bufferedInputStream)
+        val jos = ZipOutputStream(output)
+
+        // first write all non .class zip entries from the original input stream to the output stream
+        // we read it first to close the input stream as fast as possible
+        var zipEntry: ZipEntry?
+        while (jis.nextEntry.also { zipEntry = it } != null) {
```
Lucaskyy, line 55: This implementation has a flaw. There is no way to ***remove*** a class. Please look into that.

src/main/kotlin/net/revanced/patcher/util/Io.kt
```diff
@@ -3,47 +3,90 @@ package net.revanced.patcher.util
...
-                val e = jar.nextJarEntry ?: break
-                if (e.name.endsWith(".class")) {
-                    val classNode = ClassNode()
-                    ClassReader(jar.readBytes()).accept(classNode, ClassReader.EXPAND_FRAMES)
-                    this.add(classNode)
-                }
-                jar.closeEntry()
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+import java.util.zip.ZipOutputStream
```
Lucaskyy, line 11: Unused imports. Please run `Optimize Imports` in IntelliJ to remove all unused imports.

Lucaskyy (changes requested):
Looks good!

src/test/kotlin/net/revanced/patcher/PatcherTest.kt
oSumAtrIX: I an not sure if they are gonna be the same, since we rewrite into a new .jar file, so this might not make much sense

src/main/kotlin/net/revanced/patcher/util/Io.kt
oSumAtrIX: There should be, we have a List of ClassNodes, if we want to remove a class, we simply get rid it from there.

src/test/kotlin/net/revanced/patcher/PatcherTest.kt
Lucaskyy: That's true. I have an idea though, I'll checkout this PR locally and test if my idea works. If so I'll commit it to the PR.

src/main/kotlin/net/revanced/patcher/util/Io.kt
Lucaskyy: My bad, didn't read the comment. "Non class entries", this is fine.

Lucaskyy:
Don't worry about commit messages btw (^), I'll squash this PR after.

src/main/kotlin/net/revanced/patcher/util/Io.kt
Lucaskyy: Not that I know of. So this isn't an issue for now, you may want to put a FIXME or TODO above it though, so we know it's a "known issue" if it's ever going to become an issue.

Lucaskyy (approved):

Issue #9, opened 2022-03-21
Optional Signatures members


oSumAtrIX:
The method  `resolveMethod` expects a `Signature`:

https://github.com/ReVancedTeam/revanced-patcher/blob/db7d8f7908b0ede5f952ac0ddd67efe92776c650/src/main/kotlin/net/revanced/patcher/resolver/MethodResolver.kt#L57

Some members should be nullable, in case you certainly know, that these members are not required to resolve a method:
https://github.com/ReVancedTeam/revanced-patcher/blob/db7d8f7908b0ede5f952ac0ddd67efe92776c650/src/main/kotlin/net/revanced/patcher/signature/Signature.kt#L23-L29

In the case of `resolveMethod`, all members should be nullable, except the name.

Lucaskyy:
The reason I am against "nullability" is that that can cause UD and exceptions if not handled properly (which adds extra boilerplate). It would be better to use empty arrays for `opcodes` so in the case something goes haywire it just skips the for-loop instead of crashing full-on.

Lucaskyy:
Sounds good 👍

PR #10, opened 2022-03-21
fix: nullable signature members

oSumAtrIX:
RE: #9 

This commit will allow "partial" signatures. Example usage:

```kt
val showVideoAdsMethodData = cache.methods["show-video-ads"].findParentMethod(
    Signature(
        "show-video-ads-method",
        null,
        null,
        arrayOf(Type.BOOLEAN_TYPE),
        null
    )
) ?: return PatchResultError("Could not find parent method to 'show-video-ads'")
```

This example will try to use the cached method `show-video-ads` to find the actual method `show-video-ads-method`. To find this method in our example, it is only required to match for the parameters, since in our example, no other method in the same class has this parameter.

### DIFF
--- a/src/main/kotlin/net/revanced/patcher/resolver/MethodResolver.kt
+++ b/src/main/kotlin/net/revanced/patcher/resolver/MethodResolver.kt
@@ -56,7 +56,7 @@ internal class MethodResolver(private val classList: List<ClassNode>, private va
     companion object {
         fun resolveMethod(classNode: ClassNode, signature: Signature): PatchData? {
             for (method in classNode.methods) {
-                val (r, sr) = cmp(method, signature, true)
+                val (r, sr) = cmp(method, signature)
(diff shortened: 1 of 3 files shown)


Lucaskyy (approved):

Lucaskyy:
@oSumAtrIX Could you review, please?

oSumAtrIX:
Wouldn't L81 return to the inner lambda instead of the `cmp` method?

https://github.com/ReVancedTeam/revanced-patcher/blob/11e0f2ab6e844e2f28821418e6a820aa11738ed4/src/main/kotlin/net/revanced/patcher/resolver/MethodResolver.kt#L70-L81

Lucaskyy:
Good catch! And no, it doesn't. But I'll make it explicitly return to the `cmp` method to be sure.

Lucaskyy:
@oSumAtrIX All good now?

Issue #14, opened 2022-03-24
A better way of managing classes to keep memory usage low


Lucaskyy:
Currently, all classes are read from the input stream, then passed into a `ClassReader` and the returned `ClassNode` is saved in `classes` in `Cache.kt`.

**This is not optimal.** When testing the patcher with the stock YouTube APK, it uses **1300 MB** of RAM.
Remember, the patcher should run on mobile phones!
There's nothing we can do about the allocs and deallocs, that's just the downside of ASM.

To fix this issue, a breaking refactor needs to happen.
We need to read the class and resolve the methods "as we go" when reading the class files from the input stream.
This way we can save the useful classes to memory, and free the classes we don't need.

The bad things about this idea:
- to delete a class, you need to define an empty ClassNode in the list, since our list takes priority over the original JAR. we can add helper methods for this, though.
- referencing classes that are not referenced by the signatures will not work, since they are dropped

The good things about this idea:
- less memory usage! we only store the classes that are referenced by the signatures.

If you have suggestions, better ideas, or really just anything, please let me know!

oSumAtrIX:
Note, patches referencing other classes is inevitable, but deleting classes is not necessary. In that case we have to resolve all referenced classes at some point, maby a patch can return all classes which it has been using

Lucaskyy:
> Note, patches referencing other classes is inevitable

This is fine as long as these other classes have a signature belonging to them. So if PatchA references classB, then classB must have a signature in the signature array. Otherwise it will be dropped.

PR #16, opened 2022-03-31
Migrate to Dalvik patcher

### LINKED ISSUE #13: ClassWriter cannot compute stack frames for certain types
To reproduce this issue, please download the stock YouTube app from any source of your choice. I recommend the Vanced Codebucket mirror. Use dex2jar to convert the APK to JAR. Then run this code:
```kt
val testData = javaClass.getResourceAsStream("/stock.jar")!!
Patcher(testData, ByteArrayOutputStream(), emptyArray()).save()
```
This will fail with this error:
```
Exception in thread "main" java.lang.TypeNotPresentException: Type ysm not present
    at org.objectweb.asm.ClassWriter.getCommonSuperClass(ClassWriter.java:1025)
    at org.objectweb.asm.SymbolTable.addMergedType(SymbolTable.java:1202)
    at org.objectweb.asm.Frame.merge(Frame.java:1299)
    at org.objectweb.asm.Frame.merge(Frame.java:1197)
    at org.objectweb.asm.MethodWriter.computeAllFrames(MethodWriter.java:1610)
    at org.objectweb.asm.MethodWriter.visitMaxs(MethodWriter.java:1546)
    at org.objectweb.asm.tree.MethodNode.accept(MethodNode.java:769)
    at org.objectweb.asm.tree.MethodNode.accept(MethodNode.java:649)
    at org.objectweb.asm.tree.ClassNode.accept(ClassNode.java:452)
    at app.revanced.patcher.util.Io.saveAsJar(Io.kt:86)
    at app.revanced.patcher.Patcher.save(Patcher.kt:48)
    ... 2 more
Caused by: java.lang.ClassNotFoundException: ysm
    at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    at java.lang.Class.forName0(Native Method)
    at java.lang.Class.forName(Class.java:348)
    at org.objectweb.asm.ClassWriter.getCommonSuperClass(ClassWriter.java:1023)
    ... 12 more
```
This exception is thrown because the standard implementation of `getCommonSuperClass` in `ClassWriter` uses the system `ClassLoader` to find said class to get the type out of it (it will invoke `Type#getType(Class<?> clazz)`).

This issue can be mitigated by telling setting the `mode` to `COMPUTE_MAXS` instead of `COMPUTE_FRAMES`.
But this will not work when adding branches (afaik) because only the local vars are computed, and new frames for each branch are not. This means you need to add the stack frame yourself if you patch the code.
Since the patcher is also meant as a general-use library, we should not rely on the experience of the developers to write stack frames themselves. We should really fix this issue instead.
I'm not yet sure how to test this, I will have to look into adding tests for this case.

The intended solution for this is to make our own `ClassWriter` implementation and override the `getCommonSuperClass` to use our class cache and get the types out of that. Note that computing frames is really expensive and will take up some CPU and memory.

oSumAtrIX:
This PR closes #13, #14 and #17.
Due to how ASM parses class files, avoiding high memory usage is unpractical, if possible. 
For that reason, I migrated the patcher to patch Dalvik, which uses `dexlib` and way less memory.


src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -29,12 +29,13 @@ class Patcher(
...
             override fun getClasses(): Set<ClassDef> {
                 // this is a slow workaround for now
-                val mutableClassList = cache.classes.toMutableList()
+                val classes = cache.classes.toMutableSet()
                 cache.classProxy
                     .filter { it.proxyUsed }.forEach { proxy ->
-                        mutableClassList[proxy.originalIndex] = proxy.mutatedClass
+                        classes.remove(classes.elementAt(proxy.originalIndex))
+                        classes.add(proxy.mutatedClass)
                     }
```
Lucaskyy, line 37: This looks invalid. Use something like `insert` instead.

oSumAtrIX: we first remove the original class from the set of classes, then add the mutable class to the list, what's invalid about it?

Lucaskyy: You remove the original class at index `proxy.originalIndex`, then you add the modified class at index `classes.size`.

oSumAtrIX: Yeah that's the intended way

Lucaskyy: Yep, forgot about that. Order doesn't matter here.

oSumAtrIX:
Whats also missing is that "O" in signatures should match to non primitive types like Boolean, Int or Float. If possible we should also handle String as a primitive type due to it being effective when using in signatures

Lucaskyy:
So, "O" is ***anything***? (basically the same as `??` in IDA signatures.)
Then shouldn't we too just change it to `?` or `??`, to keep with the standards?

oSumAtrIX:
"O" would stand for Object, in that case we can also refer to it as "?"

oSumAtrIX:
We need to add testdex container resources for the tests to pass

Lucaskyy (approved):
Looks good!

Lucaskyy:
> @Lucaskyy, we need to add testdex container resources for the tests to pass

I think you can pull that off by getting my old `test1.jar`, using `jar2dex` (or `dx`) and then packing that into an APK (iirc they're just ZIPs, so just ZIP it). That should do the trick.

src/main/kotlin/app/revanced/patcher/extensions/Extensions.kt
```diff
@@ -1,5 +1,13 @@
...
 
 import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.builder.BuilderInstruction
+import org.jf.dexlib2.builder.MutableMethodImplementation
 
-infix fun AccessFlags.or(other: AccessFlags) = this.value or other.value
\ No newline at end of file
+infix fun AccessFlags.or(other: AccessFlags) = this.value or other.value
+
+fun MutableMethodImplementation.addInstructions(index: Int, instructions: List<BuilderInstruction>) {
```
Lucaskyy, line 9: You could make `instructions` a `vararg` instead.

oSumAtrIX: This would prevent the following:
image

Lucaskyy: You can fix that by putting `*` in front. (unpacking the array)

oSumAtrIX: Using the spread operator would only work if we would have an array instead of a list
image

baiorett (approved):

Lucaskyy:
Commit 28ed4793e39d2a2d10498f5e0b22d61f7a82dff4 closes #17.

Lucaskyy:
This PR, and the patcher in general has now entered it's final stage. The last things that need to be done, are:
* Finish the fork of dexlib2, so #20 can be closed.
* Finish class merger (assigned to @oSumAtrIX).
* Add fuzzy searcher, allowing signatures to specify a threshold and returning warnings for inaccurate signatures.

Lucaskyy:
@oSumAtrIX As for why it creates more dex files than "needed", that is due to multi-threading. I encountered it before. Should we remove multi-threading and close #17 as `wontfix`?

oSumAtrIX:
There is no issue in having more .dex files. We should favour speed where we can. Currently we do not even use multithreading iirc, so i am not sure why this should be the reason for it.

Lucaskyy:
> Currently we do not even use multithreading iirc, so i am not sure why this should be the reason for it.

We do! As you can see here:
https://github.com/ReVancedTeam/revanced-patcher/blob/2b888e381ce2df8581635622f6e71ebfea8091a7/src/main/kotlin/app/revanced/patcher/Patcher.kt#L80-L81

oSumAtrIX:
In that case we can close #17. There is no issue with multithreading currently. Different amounts of dex file container should work flawlessly

Lucaskyy:
Alright. You can let this PR close the issue automatically on merge.

Lucaskyy:
@oSumAtrIX Also, I added multi-threading yesterday as I said in https://github.com/ReVancedTeam/revanced-patcher/pull/16#issuecomment-1093462389. So that's probably why you didn't know yet it was added.

src/test/kotlin/app/revanced/patcher/PatcherTest.kt
```diff
@@ -158,6 +170,7 @@ internal class PatcherTest {
         }
 
         val out = patcher.save()
+        File("a.dex").writeBytes(out["classes.dex"]!!.data)
```
Lucaskyy, line 173: You accidentally committed this line ^ Could you remove it?

Lucaskyy: Removed in 4022b8b847e8767ace0da3f98ad72ab61a4c242b.

oSumAtrIX:
Resources have to be added to the test resource files

PR #30, opened 2022-05-07
Add `p` naming scheme to smali compiler

j4k0xb:
Makes it easier to reference the `this` register in virtual methods or parameters in general.
E.g.
```kt
// p0 (same as v2) refers to the first parameter of some static method
"return p0".toInstruction("BZI", 5)
```
Passing the method's parameter signature and register count is necessary so smali can calculate the corresponding `v` register number.


oSumAtrIX:
Thanks

PR #37, opened 2022-05-27
fix: bypass if not a class

ponces:


oSumAtrIX:
Is this tested? It looks like this will never return due to className never having an extension

tillnelown:
I gave it a try, returns java.lang.ClassNotFoundException: META-INF like without the patch.

ponces:
It was tested in terms of compilation (which not even worked before this)
I'm currently checking the binary produced...

oSumAtrIX:
the patcher compiles already, see the workflow and a commit has already been done locally fixing the related issue anyways.

ponces:
Next time, please make sure sources are always buildable.
Having broken sources and even generating packages with them doesn't make sense :(

oSumAtrIX:
The sources are buildable as already mentioned in the comment before. This is the ˋdevˋ branch and code is bound to break.

PR #51, opened 2022-06-26
refactor: migrate from `Signature` to `Fingerprint`

oSumAtrIX:
This PR migrates from using the term `Signature` to `Fingerprint`. 
It also improves the `MethodSignatureResolver` in performance and code quality.


Sculas:
This is the wrong commit type.

oSumAtrIX:
suggest a right commit message I will force push. The commit type should've been in the footer.

build.gradle.kts
```diff
@@ -21,7 +21,7 @@ repositories {
 }
 
 dependencies {
-    api("app.revanced:multidexlib2:2.5.2.r2")
+    implementation("app.revanced:multidexlib2:2.5.2.r2")
```
Sculas, line 24: How are opcodes done on the patches? Because now `Opcode.*` is inaccessible.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * Please note, this method allocates a [ClassProxy].
+     * Use [method] where possible.
+     */
+    val mutableMethod by lazy {
+        mutableClass.methods.first {
+            it.softCompareTo(this.method)
+        }
+    }
+    }
```
Sculas, line 73: I don't think this should be here. If it is, please reformat the file.

src/main/kotlin/app/revanced/patcher/fingerprint/method/utils/MethodFingerprintUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            if (stringsList.isNotEmpty()) return false
+        }
+
+        val patternScanResult = if (methodFingerprint.opcodes != null) {
+            context.implementation?.instructions ?: return false
+
+            context.patternScan(methodFingerprint) ?: return false
+            //context.patternScan(methodFingerprint) ?: return false
```
Sculas, line 94: This comment should be probably be deleted.

Sculas (changes requested):

build.gradle.kts
oSumAtrIX: The patcher is not responsible for that. It is the patch's purpose to use that dependency, note the usage of the patcher can be merely to merge integrations, and that's it where a dependency to revanced:multidexlib2  is not required. The commit is intended.

Sculas:
> suggest a right commit message I will force push. The commit type should've been in the footer.

Commit message: `refactor: migrate from Signature to Fingerprint`
Commit description: `BREAKING CHANGE: Not backwards compatible, since a lot of classes where renamed.`

build.gradle.kts
Sculas: Then people will have to update multidexlib accordingly with the patcher, which may not always be wanted. But your reasoning seems fair. Though documenting it is required. And we should refrain from updating multidexlib as much as possible then to prevent such inconveniences from happening. This is a opensource project, you should think about your fellow developers who may decide to use this.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
Sculas: If this was resolved, please resolve the conversation.

src/main/kotlin/app/revanced/patcher/fingerprint/method/utils/MethodFingerprintUtils.kt
Sculas: If this was resolved, please resolve the conversation.

Sculas (approved):

Sculas:
Make sure to squash.

oSumAtrIX:
Squashing is not required since it is a single commit :)

PR #52, opened 2022-06-27
feat: improve Smali compiler

Sculas:
- Branching support has been added. See InlineSmaliCompilerTest.kt for an example.
- Some other improvements have been made too.


Sculas:
@oSumAtrIX We need this merged. Could you start working on this now? If this won't be merged for another week I'm going to merge it since it's needed in certain patches and it's just nice to have overall.

oSumAtrIX (approved):

TheJeterLP:
I'm going to test this out later this day with my patch :)

src/test/kotlin/app/revanced/patcher/util/smali/InlineSmaliCompilerTest.kt
```diff
@@ -0,0 +1,104 @@
...
+internal class InlineSmaliCompilerTest {
+    @Test
+    fun `compiler should output valid instruction`() {
+        val want = BuilderInstruction21c(Opcode.CONST_STRING, 0, ImmutableStringReference("Test")) as BuilderInstruction
+        val have = "const-string v0, \"Test\"".toInstruction()
+        instructionEquals(want, have)
+    }
+
+    @Test
+    fun `compiler should support branching with own branches`() {
```
Sculas, line 29: Does this test still pass? If so, this can be merged.

oSumAtrIX: Yes, all tests pass. My solution does not require to fix instructions as my order of operation fixes the issue to begin with. First, compile, then insert the sublist without the dummy npps and lastly replace the labels by calcuting where in the compiled insn list the offset instructions previously pointed to.

PR #75, opened 2022-07-27
fix: close ExtFiles

danthe1st:
Resources should always be closed when they are not used any more.

This PR fixes the issue by closing `ExtFile`s created explicitly.

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/Patcher.kt
+++ b/src/main/kotlin/app/revanced/patcher/Patcher.kt
@@ -183,7 +183,6 @@ class Patcher(private val options: PatcherOptions) {
                     resources.setSparseResources(metaInfo.sparseResources)
                 }
 
-
                 val manifestFile = cacheDirectory.resolve("AndroidManifest.xml")
 
                 ResXmlPatcher.fixingPublicAttrsInProviderAttributes(manifestFile)
--- a/src/main/kotlin/app/revanced/patcher/Patcher.kt
+++ b/src/main/kotlin/app/revanced/patcher/Patcher.kt
@@ -48,70 +48,74 @@ class Patcher(private val options: PatcherOptions) {
 
     init {
         val extInputFile = ExtFile(options.inputFile)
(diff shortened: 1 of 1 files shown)


Sculas:
Could you fix the indentation?

Sculas (approved):

src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -48,70 +48,74 @@ class Patcher(private val options: PatcherOptions) {
 
     init {
         val extInputFile = ExtFile(options.inputFile)
```
oSumAtrIX, line 50: Use `.use` extension here instead of try-finally blocks.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patcher/Patcher.kt
Sculas: `ExtFile` does not implement `AutoCloseable`, so this will not work.

src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -164,43 +168,47 @@ class Patcher(private val options: PatcherOptions) {
...
+                    resources.setVersionInfo(metaInfo.versionInfo)
+                    resources.setSharedLibrary(metaInfo.sharedLibrary)
+                    resources.setSparseResources(metaInfo.sparseResources)
                 }
 
-                resources.setSdkInfo(metaInfo.sdkInfo)
-                resources.setVersionInfo(metaInfo.versionInfo)
-                resources.setSharedLibrary(metaInfo.sharedLibrary)
-                resources.setSparseResources(metaInfo.sparseResources)
-            }
```
oSumAtrIX, line 186:
```suggestion
```

oSumAtrIX (changes requested):

Issue #76, opened 2022-07-28
feat: only do resource patching when required


bogadana:
### Type

Functionality

### Issue

Currently the user has to manually specify whether resource patching should be enabled. 

### Feature

Choosing automatically based on whether any resource patches are selected removes the unnecessary manual choice.

### Motivation

Disabling resource patching when not required saves a lot of time.

bogadana:
And before you ask this can't easily be done in cli because of dependencies

Aunali321:
Same case with manager ^

oSumAtrIX:
> And before you ask this can't easily be done in cli because of dependencies

Why?

bogadana:
> > And before you ask this can't easily be done in cli because of dependencies
> 
> Why?

Are dependencies accessible without an instance?

oSumAtrIX:
Yes, you have access to the classes annotations before having to create an instance of the class.

bogadana:
Still I think it makes more sense in patcher. There's no reason to give it resource patches and then disable resource patching. Would avoid duplicate code between the cli/manager too

PR #77, opened 2022-07-28
feat: decode resources only when necessary

bogadana:
TODO: 
- First decode resources then run patches

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/Patcher.kt
+++ b/src/main/kotlin/app/revanced/patcher/Patcher.kt
@@ -1,7 +1,6 @@
 package app.revanced.patcher
 
 import app.revanced.patcher.data.Data
-import app.revanced.patcher.data.PackageMetadata
 import app.revanced.patcher.data.impl.findIndexed
 import app.revanced.patcher.extensions.PatchExtensions.dependencies
 import app.revanced.patcher.extensions.PatchExtensions.patchName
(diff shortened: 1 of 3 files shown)


bogadana:
Implements https://github.com/revanced/revanced-patcher/issues/76

Sculas (approved):

oSumAtrIX (changes requested):
The current implementation will run the patches and when a resource patch is ran, decode the resources. This will look like the patch takes long to run the patch also furthermore breaking the single responsibility principle of the method. Instead, before running the patches, check, a resource patch has been added and decode resources.

bogadana:
> This will look like the patch takes long to run the patch

no as an appropriate log message is printed

> before running the patches, check, a resource patch has been added and decode resources.

tell me how to do this without an instance

oSumAtrIX:
> no as an appropriate log message is printed

While running the patches this would be random.

bogadana:
resource patch -> resources need to be decoded
dont see how thats random

oSumAtrIX:
> While running the patches this would be random.

This is why.

bogadana:
Things should always be done as late as possible its not random at all

oSumAtrIX:
While running the patches it doesn't make sense to decode the resources at the same time. Running the patches is one thing, decoding the resources a different. Doing the latter while the patches run is random. Please decode the resources before running the patches by checking if a resource patch has been supplied.

bogadana:
then answer my other question or ditch reflection already

oSumAtrIX:
I have answered your question in the previous message why it is random and reflection will be ditched when annotations are ditched. Please apply the requested changes.

PR #84, opened 2022-08-03
feat: cross-platform patch bundle format

### LINKED ISSUE #54: feat: cross-platform patch bundle format
# Issue: 

Patches can have resources. On Android, this will not be possible. 

# Solution

Introduce a patch bundle format. Can be a ZIP file with the dex'ed resources for `DexPatchBundle` and the regular JAR file for `JarPatchBundle`.
The patcher would either need to introduce an API for patch resources or extract the resources to the patch bundle's specific cache directory to which a resource patch will have access via the `data` parameter. The implementation of the API on the other hand would require to return a  `File` for resources from the ZIP file.  Example: `data.patchBundleResources[resourcePath]`. This way writing to the cache directory would not be necessary anymore.

RE: https://github.com/revanced/revanced-patches/pull/76

Sculas:
Implements a cross-platform patch bundle format
Resolves #54


src/main/kotlin/app/revanced/patcher/util/patch/bundle/Compressor.kt
```diff
@@ -0,0 +1,34 @@
...
+    setInput(bytesIn)
+    val bytesOut = ByteArrayOutputStream(bytesIn.size).use { stream ->
+        val buffer = ByteArray(1024)
+        while (!finished()) {
+            val count = inflate(buffer)
+            stream.write(buffer, 0, count)
+        }
+        stream.toByteArray()
+    }
+    return bytesOut
```
oSumAtrIX, line 33: Avoid duplicate code here.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patcher/util/patch/bundle/Compressor.kt
Sculas: I can't. `Deflater` and `Inflater` are separate types.

oSumAtrIX: Use a generic type parameter and use it in line 28.

Sculas: I have an idea.

Sculas: Should be fixed now.

Sculas:
I've converted this to a draft because I'd like to finish the Gradle plugin I'm working on before merging this.

Sculas:
It works, but I don't like it. I feel I've done enough for this PR now so I'm going to move on to something else.
Someone else may continue with this PR and make something better out of it.

TheJeterLP:
> It works, but I don't like it. I feel I've done enough for this PR now so I'm going to move on to something else.
> Someone else may continue with this PR and make something better out of it.

So this is probably never going to get finished now lol

Issue #72, opened 2022-07-17
feat: allow defining of version ranges


TheJeterLP:
## 🐞 Issue

Currently we provide versions by a List containing strings, and the Patcher uses those. So for every new version we have to add another string to the list. We should find a way to be able to create a version range (for ex. 17.24.35-17.29.36). This will cleanup the source of Patches by a lot and makes it easier to maintain. 

## ❗ Solution

Get version string, replace "." With nothing, so the version will look like 172435 and convert that to an int. Then we can check if the provided APK is in the specific version range.

## ❓ Motivation

Cleanup of code, not having to list every minor version for compatibility

oSumAtrIX:
Is the solution proposal fail safe? Something like 1.2.3 would fail against 1.12.34

j4k0xb:
I'd suggest using the version parser from the patches (https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/meta/readme/SemanticVersion.kt) to compare them

TheJeterLP:
> Is the solution proposal fail safe? Something like 1.2.3 would fail against 1.12.34

With YouTubes current versioning system it is. If they change that, of course it won't be any longer.

TheJeterLP:
> I'd suggest using the version parser from the patches (https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/meta/readme/SemanticVersion.kt) to compare them

If that works, no problem yes.

oSumAtrIX:
Because `versionName` is used, an arbitary string can be used for versions, which is why by that generalization a range can not be implemented by design.

Ushie:
What about supporting regex expressions?

PR #94, opened 2022-09-01
chore: copy the semantic version comparator to the patcher

MewtR:
Looking at #72, I figured it would be best to move the semantic version comparator to the patcher. It can then be imported in the cli as well revanced-patches.


oSumAtrIX:
It can not yet close #72. Also it should be removed from the CLI. Additionally this is not really the patchers job, the CLI is who checks for the version. The plan is to create a shared library between the manager and the CLI, so both the CLI and the manager can then make use of shared functions.

MewtR:
I'm aware it doesn't fully implement #72. I put it in the patcher purely because the cli and revanced-patches both depend on the patcher, to avoid having duplicate code. But I agree it should be the cli's job to do the version checking (as is already the case).

PR #111, opened 2022-09-29
fix: remove useless dummy nops

### LINKED ISSUE #91: bug: `MutableMethod.addInstructions` does not remove dummy nops
### Type

Other

### Bug description

When the `externalLabels` parameter is used, dummy nops are generated to create temporal offsets. These do not get removed when the instructions are added to the method.

### Steps to reproduce

Call `MutableMethod.addInstructions` with the `externalLabels` parameter. Observe the inserted instructions

## Usage:

image

## Result:

image

### Solution

Remove the dummy nops after they are not used anymore.

MewtR:
Closes #91


src/main/kotlin/app/revanced/patcher/extensions/Extensions.kt
```diff
@@ -177,6 +178,14 @@ fun MutableMethod.addInstructions(index: Int, smali: String, externalLabels: Lis
             // this line works because we created the dummy instructions in the same order as the externalLabels list.
             val (_, instruction) = externalLabels[(compiledInstructions.size - 1) - labelIndex]
             instruction.makeNewLabel()
+
+            // Remove the dummy nop, the following instruction.
```
oSumAtrIX, line 182:
```suggestion
            // remove the dummy nops
```

oSumAtrIX (changes requested):

oSumAtrIX:
The method creates multiple dummy nops, this PR would only consider one, is this considered?

MewtR:
Yeah you're right, I updated it so that it takes care of multiple nops if necessary.

oSumAtrIX:
I am not quite happy with the current implementation, isn't it possible to get rid of the nops with the current information without having to keel track of the indices?

MewtR:
I'll look into it

MewtR:
Something like this would work.

src/main/kotlin/app/revanced/patcher/extensions/Extensions.kt
```diff
@@ -177,6 +180,21 @@ fun MutableMethod.addInstructions(index: Int, smali: String, externalLabels: Lis
             // this line works because we created the dummy instructions in the same order as the externalLabels list.
             val (_, instruction) = externalLabels[(compiledInstructions.size - 1) - labelIndex]
             instruction.makeNewLabel()
+
+            // Since we delete the nops as we're iterating, to find the index of subsequent
+            // nops we need to substract the number of nops already removed
+            val nopInstructionIndex = index + labelIndex - nopInstructionsRemoved;
+
+            // Remove the corresponding nop.
+            if (nopInstructionsRemoved < externalLabels.size)
```
oSumAtrIX, line 189: Instead of this, why not just add the compiled instructions without the nops to the target method implementation instead of adding the entire instruction list, and then removing the nop instructions from them? Above the instructions are compiled, you could instead just add a subList of the instructions without the nops.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patcher/extensions/Extensions.kt
MewtR: Something like this?

oSumAtrIX: I think so, is this tested?

MewtR: Yeah it works, but I think it looks a little awkward.

Sculas (approved):

oSumAtrIX (approved):

Issue #73, opened 2022-07-18
feat: split support


Sculas:
## 🐞 Issue
Split support is missing, meaning some patches don't work with the root method.

## ❗ Solution
Add split support.

## ❓ Motivation
The root method does not work with some patches.

## ⚠ Additional context
This issue will be improved later on.

Guigeek973:
Hello ! 
oSumAtrIX removed this from [Development of ReVanced] [on May 19]
do that mean that it's not anymore planned to add armv7 devices ?

oSumAtrIX:
Split support has nothing to do with arm and this is superseded by arsclib.

masterflitzer:
may i ask when arsclib will be implemented? i couldn't find anything in the gh issue tracker

oSumAtrIX:
No ETA.

AndnixSH:
I have a suggestion. Implement APKeditor to merge apk into single apk then finally it can patch single apk. I have been using APKeditor a lot for APK modding purposes, and it's working great without issues

Repo: https://github.com/REAndroid/APKEditor

yonggamer:
[APKEditor](https://github.com/REAndroid/APKEditor) [V1.3.7](https://github.com/REAndroid/APKEditor/releases/tag/V1.3.7)
gives problems for me with [YouTube 19.09.37](https://www.apkmirror.com/apk/google-inc/youtube/youtube-19-09-37-release/youtube-19-09-37-android-apk-download/) [revanced-cli](https://github.com/ReVanced/revanced-cli) [v4.6.0](https://github.com/ReVanced/revanced-cli/releases/tag/v4.6.0) [revanced-patches](https://github.com/ReVanced/revanced-patches) [v4.6.0](https://github.com/ReVanced/revanced-patches/releases/tag/v4.6.0) [revanced-integrations](https://github.com/ReVanced/revanced-integrations) [v1.7.0](https://github.com/ReVanced/revanced-integrations/releases/tag/v1.7.0)

yonggamer:
https://github.com/ReVanced/revanced-manager/issues/1511#issuecomment-2045799258

Th3-A6add0n:
> I have a suggestion. Implement APKeditor to merge apk into single apk then finally it can patch single apk. I have been using APKeditor a lot for APK modding purposes, and it's working great without issues
> 
> Repo: https://github.com/REAndroid/APKEditor

[...]

PR #113, opened 2022-09-30
feat: split apk files

### LINKED ISSUE #104: feat: `ApkPatch`
### Type

Functionality

### Issue

Afaik, there are only 2 patch(BytecodePatch and ResourcePatch) at the moment. And ResourcePatch will not work if there is a resource decode problem with apktool. So with ApkPatch(or other name suitable) can at least help add raw resources to `assets` in some case.

### Feature

It is a patch can edit the `apk` by add, remove or edit the file in `apk` like file in `assets`, `lib`, ...

### Motivation

I see some people complain about the `universal apk` size. So which this `ApkPatch` it help to add a patch to reduce the `apk` size by remove unnecessary file in `apk` like not target `shared library`. In addition, it can also help add raw resources to `assets` if there is a patch need it and the apk's resources can not be decoded.

### Additional context

_No response_

oSumAtrIX:
## About

- This PR adds support for split apks. The API usage can be seen on the implementing CLI branch: https://github.com/revanced/revanced-cli/tree/feat/split

- Closes #104, #73.

## TODO

- [ ] Test the PR
- [ ] Implement TODOs


src/test/kotlin/app/revanced/patcher/usage/resource/patch/ExampleResourcePatch.kt
```diff
@@ -18,6 +18,7 @@ import org.w3c.dom.Element
 @Version("0.0.1")
 class ExampleResourcePatch : ResourcePatch() {
     override fun execute(data: ResourceData): PatchResult {
+        data
```
Sculas, line 21:
```suggestion
```

Sculas (changes requested):

oSumAtrIX:
@Sculas the problem with just generically detecting whether it is a split or not is, that after patching it is required to know which apk is what, the language, assets or other apks. Now one might say, why not supply one apk each to the patcher, the problem with that is, that a patch would not be able to access all apks at the same time, like it does currently with a full apk, which is why I went with a sealed Apk class and a defined set of Subclasses for the splits.

Sculas:
I understand, but all Java code is only ever in the base APK. All other splits do not contain dex files. So the problem you're trying to solve never existed to begin with.

Sculas:
Either way, if this implementation works then it's good enough. Improvements can be made with time.

bao000 (commented):
<dependency>
  <groupId>app.revanced</groupId>
  <artifactId>revanced-patcher</artifactId>
  <version>6.3.0</version>
</dependency>

oSumAtrIX:
Superseded by #165.

PR #115, opened 2022-10-10
fix: slightly better fuzzy scanning

### LINKED ISSUE #89: bug: fuzzy scanner failing if instructions are missing
### Type

Other

### Bug description

If a pattern of a fingerprint is missing an opcode or has more opcodes than expected, the fuzzy scan will fail. Currently the fuzzy scan only allows _wrong_ opcodes only.

### Steps to reproduce

1. Create a fingerprint with a reasonable opcode pattern size
2. Fuzzy scan it
3. Change one opcode and confirm the fuzzy scanner works as expected
4. Remove an opcode  and confirm the fuzzy scanner now fails

### Relevant log output

```shell
-
```

### Screenshots or videos

_No response_

### Solution

A proper fuzzy scanning algorithm should be used instead of the current barebone implementation.

### Additional context

_No response_

MewtR:
Closes #89


src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -158,18 +158,33 @@ abstract class MethodFingerprint(
...
-                    val patternOpcode = pattern.elementAt(patternIndex)
+                    val patternOpcode = pattern.elementAt(patternIndex + offset)
 
                     if (patternOpcode != null && patternOpcode.ordinal != originalOpcode.ordinal) {
                         // reaching maximum threshold (0) means,
                         // the pattern does not match to the current instructions
                         if (threshold-- == 0) break
+
+                        // look ahead
+                        for (depth in 1 until threshold + 2) {
```
oSumAtrIX, line 173: Is `threshold + 2` arbitrary?

MewtR: Kind of, I wanted to make it so that at a minimum at least the next instruction is considered. If threshold is 0 at that point, you'll only enter that loop once since you start at 1 and would go 'until' 2 (exclusive).

Nitroandaccounts: @oSumAtrIX

oSumAtrIX:
Before merging, can you rebase & test this PR against all patches for YouTube?

Sculas:
Can this be merged soonish? PR has been open for a long while now, waiting for @oSumAtrIX to review.

oSumAtrIX:
If possible, please test this against a majority of the existing patches.

MewtR:
Hey, apologies for the delayed reply. I won't be able to look at this for quite some time. If someone else can pick it up that'd be great or else I think it'd be safer to close the PR.

oSumAtrIX:
Well it can stay open just fine until we can test it.

Sculas:
@oSumAtrIX This should be reopened.

Sculas:
What's keeping this PR from being merged after almost a year? Is no one available from the team to test this change, if necessary?
This PR is quite essential, given that my implementation of fuzzy scanning is pretty mediocre.

oSumAtrIX:
It has low priority because it is currently unnecessary

oSumAtrIX:
I believe, if we want to properly fix and implement fuzzy scanning, something like https://github.com/intuit/fuzzy-matcher should be used instead.

PR #116, opened 2022-10-13
fix: make `classes` property public

Canny1913:


oSumAtrIX:
Why is it required to be public?

Canny1913:
Because CLI cannot access this class without it being public.

oSumAtrIX:
I don't think the CLI should have access.to the class list. When does it need access?

Canny1913:
https://github.com/revanced/revanced-cli/blob/feat/split/src/main/kotlin/app/revanced/utils/patcher/Patcher.kt#L17

PR #123, opened 2022-11-13
refactor!: cleanup patch options

Sculas:
This PR cleans up the patch options system to only what's needed, simplifies it a lot, and improves it overall.
Things have been moved into their respective package as well, so this is a breaking change.

This PR also includes a small general interface for serialization, but it can be removed if requested.


src/main/kotlin/app/revanced/patcher/patch/options/OptionsContainer.kt
```diff
@@ -0,0 +1,86 @@
...
+    ): PatchOption<T> {
+        val option = PatchOption(key, title, description, defaultValue)
+        if (registry.containsKey(option.key)) {
+            throw IllegalStateException("Multiple options found with the same key")
+        }
+        registry[option.key] = option
+        return option
+    }
+
+    protected fun stringPreference(
```
Sculas, line 45: @oSumAtrIX What naming do you prefer?
`stringPreference` or `stringOption`?

oSumAtrIX (changes requested):

oSumAtrIX:
Why has this been closed?

Sculas:
Forgot I had this PR open, so I closed it since I don't have motivation to work on this PR right now (working on other things).

PR #129, opened 2022-12-07
feat: action for dev2main

Sculas:
As requested by @oSumAtrIX.

I have not tested this, because GitHub does ***not*** allow me to. The actions tab stays empty.
But... looking at the code it should be fine.


oSumAtrIX:
Moved to correct branch ,https://github.com/revanced/revanced-patcher/pull/130.

PR #160, opened 2023-02-23
Update Gradle wrapper to 8.0.1

JavierFlores09:
And also ignore gradle.properties, we don't want to share the github token.

This change is necessary to compile the project with Java 19. Gradle 8 doesn't introduce any breaking changes to the workflow as far as I've tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,6 @@ gradle-app.setting
 
 # Dependency directories
 node_modules/
+
+# Gradle props, to avoid sharing the gpr key
+gradle.properties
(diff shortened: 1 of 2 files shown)


JavierFlores09:
> Can this change be reflected in other repositories as well? Is the third requirement [here](https://github.com/revanced/revanced-cli/blob/main/docs/0_prerequisites.md#-requirements) considered as well?

Yeah, I can do that. As for the requirement, you can still compile with JDK 17 just fine, this only raises the compatibility of JDK versions (aside from going from Gradle 7->8 which may introduce some features worth looking forward); if you want me to update the requirement however I can do that too.

oSumAtrIX (approved):

PR #167, opened 2023-05-04
fix(extensions): Fix off-by-one in removeInstructions

rstular:


src/main/kotlin/app/revanced/patcher/extensions/Extensions.kt
```diff
@@ -39,7 +39,7 @@ fun MutableMethodImplementation.replaceInstructions(index: Int, instructions: Li
 }
 
 fun MutableMethodImplementation.removeInstructions(index: Int, count: Int) {
-    for (i in count downTo 0) {
+    for (i in (0 until count).reversed()) {
```
oSumAtrIX, line 42: Why not just:
```suggestion
    for (i in count - 1 downTo 0) {
```

oSumAtrIX (changes requested):

Sculas:
@oSumAtrIX Should be reopened as well.

Sculas (approved):

PR #185, opened 2023-05-31
perf: resolve fingerprints using method maps

LisoUseInAIKyrios:
Instead of looking thru every single method for each method signature, instead selectively check only the class methods that match the signatures:
- access type
- return type
- parameter types
- or contain a declared strings

Time to patch YouTube:
before:  54 seconds
after (this PR): 46 seconds

This is approaching the speed of the [one off branch](https://github.com/LisoUseInAIKyrios/revanced-patcher/tree/cached_resolving) I made that serializes the resolved class names to json (which gives patch time of 38 seconds).  But getting that performance requires patching multiple times locally, or bundling an additional resolver json file with the patches (which isn't ideal).

The changes of this PR work without any extra data, and simply cull the number of methods required to check.

Further performance improvements could be made in the patches by adding the access and return type to more method signatures.  I fixed the slowest resolving fingerprints in [this branch](https://github.com/LisoUseInAIKyrios/revanced-patches/tree/resolving_performance_tweaks), and using those changes with this patcher modifications the time to patch is reduced to 40 seconds.

Not a big deal if this is never merged, as I'll keep using the [one off branch](https://github.com/LisoUseInAIKyrios/revanced-patcher/tree/cached_resolving) which gives even faster performance.  But this PR might be useful for anyone else.


src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -39,6 +42,7 @@ class Patcher(private val options: PatcherOptions) {
     private var resourceDecodingMode = ResourceDecodingMode.MANIFEST_ONLY
     private var mergeIntegrations = false
     val context: PatcherContext
+    private lateinit var methodMap : Map<String, List<Pair<ClassDef, Method>>>
```
oSumAtrIX, line 45: This field shouldn't be present here and instead somewhere in the fingerprint classes, because it is used there

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -42,6 +44,93 @@ abstract class MethodFingerprint(
...
+                    return ""
+                }
+            }
+
+            return buildString {
+                append(accessFlags)
+                append(returnTypeValue.first())
+                parameters?.forEach { append(it.first()) }
+            }
+        }
```
oSumAtrIX: Dexlib already includes a utility class, "MethodUtils" to get the signature of a method.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -42,6 +44,93 @@ abstract class MethodFingerprint(
...
+
+        internal fun createMethodLookupMap(classes: Iterable<ClassDef>) : Map<String, List<Pair<ClassDef, Method>>> {
```
oSumAtrIX, line 70: The return type is not readable. It is not apparent by directly looking at the return value what this method is returning. Create classes and/or typealiases

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -42,6 +44,93 @@ abstract class MethodFingerprint(
...
+                            val string = ((instruction as ReferenceInstruction).reference as StringReference).string
+                            addMethodToMapList(methodMap, "String:$string", classDef, method)
+                        }
+                    }
+                }
+            }
+            return methodMap
+        }
+
+        fun Iterable<MethodFingerprint>.resolve(context: BytecodeContext, methodMap: Map<String, List<Pair<ClassDef, Method>>>) {
```
oSumAtrIX, line 112: Where is this function used? I can not find any reference

oSumAtrIX, line 112: This extension function particularly extends `Iterable<MethodFingerprint>` only; this is arbitrary, instead, functions for resolving a single MethodFingerprint should exist.  Ideally, these extension functions are moved as functions in `MethodFingerprint` instead of extension functions.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -42,6 +44,93 @@ abstract class MethodFingerprint(
...
+                    }
+
+                    // For signatures with no access or return type specified
+                    addMethodToMapList(methodMap, "", classDef, method)
+                    // Only access and return type
+                    addMethodToMapList(methodMap, accessFlagsReturnKey, classDef, method)
+                    // Access, return, and parameter count
+                    addMethodToMapList(methodMap, accessFlagsReturnParametersKey, classDef, method)
+                    // Using more fine grained lookup is possible, but in practice it does not improve performance
+
```
oSumAtrIX, line 98: I don't know why the map is designed this way. A single map is used for everything and is missing some proper structure/abstraction. Instead, a lookup table can be created for each type of key.

oSumAtrIX (commented):
All in all an interesting idea that can indeed help with speed. I am currently missing better abstractions and object-oriented approaches. To avoid this abstraction, it seems like a single map was used as a lookup table for all sorts of keys, partially arbitrary ones such as the "String:" prefix.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
oSumAtrIX: That sounds good. I am not yet entirely sure on how certain abstractions should look like for this PR, but this should work to proceed for now.

oSumAtrIX:
I'll have to review this regarding code quality and style before merging, so that we don't have to look back to it to refactor if necessary

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -40,6 +58,136 @@ abstract class MethodFingerprint(
...
+        }
+
+        internal fun clearMethodLookupMap() {
+            allMethods.clear()
+            signatureMap.clear()
+            stringMap.clear()
+        }
+
+        /**
+         * resolve faster, by creating culled lists based on method signature and Strings contained.
```
oSumAtrIX, line 130: What is "culled" lists?

oSumAtrIX, line 130: Comments should start with uppercase. "Strings" should be lowercase

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -54,6 +202,30 @@ abstract class MethodFingerprint(
                         break@classes // if the resolution succeeded, continue with the next fingerprint
         }
 
+        /**
+         * Resolve using map built in [createMethodLookupMap]
+         */
+        internal fun MethodFingerprint.resolveUsingLookupMap(context: BytecodeContext, logger : Logger = NopLogger): Boolean {
```
oSumAtrIX, line 208: Are the `resolveUsingLookupMap` extension functions even used anywhere in this project? They include a logger parameter but other resolve methods don't, this should be consistently applied to all or no extension function here.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -40,6 +58,136 @@ abstract class MethodFingerprint(
...
+        /**
+         * Resolve using map built in [createMethodLookupMap]
+         *
+         * @param logger optional logger, to record the time to resolve each fingerprint.
+         */
+        internal fun Iterable<MethodFingerprint>.resolveUsingLookupMap(context: BytecodeContext, logger : Logger = NopLogger) {
+            if (allMethods.isEmpty()) throw PatchResultError("lookup map not initialized")
+
+            for (fingerprint in this) {
+                var time = System.currentTimeMillis()
```
oSumAtrIX, line 184: This extension function measures the time, but no other does and thus introduces inconsistency across the extension functions. Either consistently measure the resolution time in every resolution function or in none.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -40,6 +58,136 @@ abstract class MethodFingerprint(
...
+
+                    // The only additional lookup that could benefit, is a map of the full class name to its methods.
+                    // This would require adding a 'class name' field to MethodFingerprint,
+                    // as currently the class name can be specified only with a custom fingerprint.
+                }
+            }
+        }
+
+        /**
+         * Resolve using map built in [createMethodLookupMap]
```
oSumAtrIX, line 176: Comments should end with punctuation and be properly worded:

```suggestion
         * Resolve using the lookup map built by [createMethodLookupMap].
```

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -40,6 +58,136 @@ abstract class MethodFingerprint(
...
+                        appendSignatureKeyParameters(method.parameterTypes)
+                    }
+                    val classAndMethod = ClassAndMethod(classDef, method)
+
+                    // For signatures with no access or return type specified.
+                    allMethods += classAndMethod
+                    // Access and return type.
+                    addMethodToMapList(signatureMap, accessFlagsReturnKey, classAndMethod)
+                    // Access, return, and parameters.
+                    addMethodToMapList(signatureMap, accessFlagsReturnParametersKey, classAndMethod)
```
oSumAtrIX, line 158: Currently, the `MethodFingerprint` API does not suggest the user design the fingerprint according to this arbitrary key structure because it is not apparent why it exists in this form and shape. The API has to be refactored so it is clear to the user. I am not sure how that should be accomplished, considering this is a highly abstract and arbitrary way of increasing performance.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -359,6 +364,7 @@ class Patcher(private val options: PatcherOptions) {
                     if (stopOnError && patchResult.isError()) return@sequence
                 }
             } finally {
+                MethodFingerprint.clearMethodLookupMap() // No more resolving can occur.
```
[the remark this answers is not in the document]
oSumAtrIX: Usually this is the garbage collectors job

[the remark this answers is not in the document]
oSumAtrIX: The map itself should be garbage collected when not used anymore.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -40,6 +58,136 @@ abstract class MethodFingerprint(
...
+                // Only check the first String declared
+                return stringMap[strings.first()]
+            }
+            return null
+        }
+
+        /**
+         * @return all app methods that could match this signature.
+         */
+        private fun MethodFingerprint.appMethodsWithSameSignature() : List<ClassAndMethod> {
```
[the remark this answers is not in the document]
oSumAtrIX: Something like suggested [here](https://github.com/revanced/revanced-patcher/pull/185#discussion_r1225640470) can be used

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
oSumAtrIX: > I think the logger should be a field that's set by another class (ie Patcher),

Currently, we use extension functions instead of simply adding the functions right to the `MethodFingerprint` class. Eventually, this will happen by a refactor, and since the functions will not be static, a logger can only be injected as a parameter, or it has to be static somewhere, but the latter is not elegant. For now, removing the logger completely should be fine, as logging in itself is worth a PR.

src/main/kotlin/app/revanced/patcher/Patcher.kt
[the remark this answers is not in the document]
oSumAtrIX: I am not sure how the garbage collector works in detail, but in this case, is it necessary to save the map in a companion object?

PR #189, opened 2023-06-11
feat: respecting order of insertions using new extension functions 

1fexd:
I have created a "concept" for new utility methods: 

* `insertInstructions` extension method on `MutableMethod` adds a simple way to insert multiple instructions without having to mange their index yourself

For example, if I want to insert two instructions at the beginning of a method, I could do this:

```kt
addInstructions(0, "instruction-1")
addInstructions(1, "instruction-2")
```

This works fine and is quite readable, but consider this:

```kt
addInstructions(0, """
    instruction-1
    instruction-2
    instruction-3
""")
addInstructions(3, "instruction-4")
```

In order to insert `instruction-4` after the first 3 instructions, I have to calculate it's index like this: `startIndex + amount_previously_added_instructions` (e.g. `0 + 3`), which might become a bit tedious.

Another option to achieve the desired result would be the following:

```kt
addInstructions(0, "instruction-4")
addInstructions(0, """
    instruction-1
    instruction-2
    instruction-3
""")
```

However, this might be confusing at first sight since one could assume the result would be this:
```
instruction-4
instruction-1
instruction-2
instruction-3
```

Therefore, I have added the `insertInstructions` which automatically calculates the index for the next instruction insert

* `hasInstruction` checks if at least one instruction has the specified `opcode` and is the correct implementation of `Instruction` and returns a `CustomFingerprint` 
* `ReferencedMethodDocumentation` is an annotation that's only present in source code and allows the developer to specify exactly which method a fingerprint targets (and in which version of the app), so it can easily be found by the developer while debugging or updating a patch
* `ImmutableExtensions` adds a `match` extension method on `ImmutableMethodReference` which allows to check if a `ReferenceInstruction` matches a `ImmutableMethodReference` (which could be used as a constant in the fingerprint)
  * This should also be added for other `Immutable*` classes to allow matching of other `Instruction`s

To demostrate the use of these utilities, I have updated the patch from [#2386 @ revanced-patches](https://github.com/revanced/revanced-patches/pull/2386) here: [BadgeTabPatch.kt](https://github.com/1fexd/revanced-patches/blob/efea59313c47193d257065ea83deaa867171ad73/src/main/kotlin/app/revanced/patches/songpal/badge/patch/BadgeTabPatch.kt#L29), [ShowNotificationFingerprint.kt](https://github.com/1fexd/revanced-patches/blob/efea59313c47193d257065ea83deaa867171ad73/src/main/kotlin/app/revanced/patches/songpal/badge/fingerprints/ShowNotificationFingerprint.kt), [CreateTabsFingerprint.kt](https://github.com/1fexd/revanced-patches/blob/efea59313c47193d257065ea83deaa867171ad73/src/main/kotlin/app/revanced/patches/songpal/badge/fingerprints/CreateTabsFingerprint.kt)

Let me know what you think.


gradle.properties
oSumAtrIX, line 1: Do not commit to this file, it is automatically updated

src/main/kotlin/app/revanced/patcher/extensions/ImmutableExtensions.kt
```diff
@@ -0,0 +1,30 @@
+package app.revanced.patcher.extensions
+
+import org.jf.dexlib2.iface.instruction.ReferenceInstruction
+import org.jf.dexlib2.iface.reference.MethodReference
+import org.jf.dexlib2.immutable.reference.ImmutableMethodReference
+import java.util.*
+
+object ImmutableExtensions {
```
oSumAtrIX, line 8: This may be moved over to the patches repository as an extension method

src/main/kotlin/app/revanced/patcher/extensions/ImmutableExtensions.kt
```diff
@@ -0,0 +1,30 @@
...
+
+        companion object {
+            val all: EnumSet<MethodReferenceMatch> = EnumSet.allOf(MethodReferenceMatch::class.java)
+        }
+    }
+
+    fun ImmutableMethodReference.matches(
+        instruction: ReferenceInstruction,
+        match: EnumSet<MethodReferenceMatch> = MethodReferenceMatch.all
+    ): Boolean {
```
oSumAtrIX, line 20: How about inverting this and doing the following:

`myReferenceInstruction.referenEquals(otherReference)`

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -15,6 +15,21 @@ import org.jf.dexlib2.util.MethodUtil
 
 private typealias StringMatch = MethodFingerprintResult.MethodFingerprintScanResult.StringsScanResult.StringMatch
 private typealias StringsScanResult = MethodFingerprintResult.MethodFingerprintScanResult.StringsScanResult
+typealias CustomFingerprint = ((methodDef: Method, classDef: ClassDef) -> Boolean)
```
oSumAtrIX, line 18: Can't this be private?

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -15,6 +15,21 @@ import org.jf.dexlib2.util.MethodUtil
...
+): CustomFingerprint {
+    return { methodDef, _ ->
+        methodDef.implementation?.instructions?.any {
+            if (it.opcode != opcode) return@any false
+            if (it !is T) return@any false
+
+            predicate(it)
+        } ?: false
+    }
+}
```
oSumAtrIX, line 32: This should probably be moved to patches

src/main/kotlin/app/revanced/patcher/fingerprint/method/annotation/MethodFingerprintMetadata.kt
```diff
@@ -9,4 +9,14 @@ import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
...
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class ReferencedMethodDocumentation(
+    val version: String,
+    val className: String,
+    val methodName: String,
+    val parameterTypes: Array<String> = [],
+    val returnType: String
+)
```
oSumAtrIX, line 22: What is this and what use case does it have?

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
```diff
@@ -11,7 +12,23 @@ import org.jf.dexlib2.builder.MutableMethodImplementation
...
+    fun putInstructions(instructions: List<BuilderInstruction>) {
+        method.implementation!!.addInstructions(index, instructions)
+        index += instructions.size
+    }
+}
+
 object InstructionExtensions {
+    fun MutableMethod.insertInstructions(startIndex: Int, fn: InstructionInserter.() -> Unit): Int {
+        return InstructionInserter(this, startIndex).apply(fn).index
+    }
```
oSumAtrIX, line 31: To adhere to consistency, this extension function should be added for the following signatures, just like all the other extension functions:

- `MutableMethod.insertInstructions(startIndex: Int, fn: InstructionInserter.() -> Unit)`
- `MutableMethod.insertInstructions(fn: InstructionInserter.() -> Unit)`
- `MutableMethodImplementation.insertInstructions(startIndex: Int, fn: InstructionInserter.() -> Unit)`
- `MutableMethodImplementation.insertInstructions(fn: InstructionInserter.() -> Unit)`

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
```diff
@@ -11,7 +12,23 @@ import org.jf.dexlib2.builder.MutableMethodImplementation
...
 import org.jf.dexlib2.iface.instruction.Instruction
 
+
+class InstructionInserter(val method: MutableMethod, startIndex: Int) {
+    var index = startIndex
+    fun putInstructions(smaliInstructions: String) {
+        putInstructions(smaliInstructions.toInstructions(method))
+    }
+
+    fun putInstructions(instructions: List<BuilderInstruction>) {
```
oSumAtrIX, line 22: Why not `addInstructions`?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
1fexd: Adding an extension method on `MutableMethodImplementation` won't be possible since in `InstructionInserter` we have a call to `String.toInstructions` which requires a `MutableMethod` instance

I have added a default value for `startIndex` to `MutableMethod.insertInstructions`, is that fine or do you prefer a second method?

src/main/kotlin/app/revanced/patcher/fingerprint/method/annotation/MethodFingerprintMetadata.kt
1fexd: > `ReferencedMethodDocumentation` is an annotation that's only present in source code and allows the developer to specify exactly which method a fingerprint targets (and in which version of the app), so it can easily be found by the developer while debugging or updating a patch

Used like this: https://github.com/1fexd/revanced-patches/blob/efea59313c47193d257065ea83deaa867171ad73/src/main/kotlin/app/revanced/patches/songpal/badge/fingerprints/ShowNotificationFingerprint.kt#L16

oSumAtrIX: And what is it used for? Why would a comment not suffice?

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
oSumAtrIX: > I have added a default value for startIndex to MutableMethod.insertInstructions, is that fine or would you prefer a second method?

No other extension method in this class uses a default value out of the reason that you would have to specify the second parameter and out of the reason because `List.add` doesn't do that either and introduces a second overload.

1fexd: Alright, done.

src/main/kotlin/app/revanced/patcher/fingerprint/method/annotation/MethodFingerprintMetadata.kt
1fexd: It isn't used at all, I just thought it would be a good way to standardize this (and a comment just looks ugly in my opinion)

gradle.properties
oSumAtrIX, line 1: Do not commit to this file or delete it; it is required.

src/main/kotlin/app/revanced/patcher/fingerprint/method/annotation/MethodFingerprintMetadata.kt
oSumAtrIX: Annotations with source retention are used for different purposes. They are used to generate code or are used by the compiler. Using such an annotation for the purpose of comments should be avoided.

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
```diff
@@ -11,7 +12,28 @@ import org.jf.dexlib2.builder.MutableMethodImplementation
...
+    }
+
+    fun addInstructions(instructions: List<BuilderInstruction>) {
+        method.implementation!!.addInstructions(index, instructions)
+        index += instructions.size
+    }
+}
+
 object InstructionExtensions {
+    fun MutableMethod.insertInstructions(startIndex: Int, fn: InstructionInserter.() -> Unit): Int {
```
oSumAtrIX, line 30: Please add comments following the style of the others and move the functions to the correct place.

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
```diff
@@ -11,7 +12,28 @@ import org.jf.dexlib2.builder.MutableMethodImplementation
 import org.jf.dexlib2.builder.instruction.*
 import org.jf.dexlib2.iface.instruction.Instruction
 
+
+class InstructionInserter(val method: MutableMethod, startIndex: Int) {
```
oSumAtrIX, line 16: I am not sure if this approach is good. Instead of allowing any extension function to be used from `InstructionExtensions`, you constrain it to two functions inside `InstructionInserter.` Instead, you should allow using any function from `InstructionExtensions.` I am not sure how this can be achieved through using function parameters.

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
```diff
@@ -19,6 +19,7 @@ import java.util.*
 private typealias StringMatch = MethodFingerprintResult.MethodFingerprintScanResult.StringsScanResult.StringMatch
 private typealias StringsScanResult = MethodFingerprintResult.MethodFingerprintScanResult.StringsScanResult
 private typealias MethodClassPair = Pair<Method, ClassDef>
+private typealias CustomFingerprint = ((methodDef: Method, classDef: ClassDef) -> Boolean)
```
oSumAtrIX, line 22: Why is an alias present for this again?

src/main/kotlin/app/revanced/patcher/extensions/InstructionExtensions.kt
1fexd: What do you mean by that?

src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt
1fexd: Because I think it makes the code more readable, and a typealias allows us to easily re-use it across the codebase.

oSumAtrIX: The alias `CustomFingerprint` is only used in this class & it should be called differently.

1fexd: Yes, but in my other PR over at revanced-patcher I added the following:

https://github.com/revanced/revanced-patches/pull/2391/files#diff-22fcba02af7ed74fb20b41d162ed1daf9d95387c1932e4ee691f45b0e13b0f06

revanced-patcher/src/main/kotlin/app/revanced/patcher/fingerprint/method/annotation/MethodFingerprintMetadata.kt
```diff
@@ -9,4 +9,14 @@ import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 @Target(AnnotationTarget.CLASS)
 annotation class FuzzyPatternScanMethod(
     val threshold: Int = 1
-)
\ No newline at end of file
+)
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class ReferencedMethodDocumentation(
```
oSumAtrIX, line 16: Whats this annotation for?

oSumAtrIX:
While an alias is a good idea to give complex types a name, it is not really necessary here. The PR seems to not have any relevant changes and as far as I am concerned can be closed. Thanks anyways.

PR #198, opened 2023-07-01
fix: clear method lookup maps

LisoUseInAIKyrios:
Since the patcher stays in memory, patching in Manager multiple times (without restarting the app) always fails on the second patch.   The lookup maps to be cleared after patching finishes.

This fix should be pushed out now, and Manager needs another dependency update and release.

manager multiple patch error


oSumAtrIX:
A proper deconstructor does not exist yet for ReVanced Patcher. Currently, ReVanced Patcher is "use-once". After you supply patches and patches, the patches can not be reused to patch again because they keep their state. The entire class loader used to load the classes has to be deleted, and a new one has to be used. A similar state issue exists with ReVanced Patcher.

PR #210, opened 2023-08-03
fix: clear `FingerprintResolutionLookupMaps` before patching

TheAabedKhan:
Right now, if we abort the patching process after merging integrations, we need to restart the app to start another patching process.

This PR aims to clear the maps from previous patching session so that new session can be started at any time.

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/Patcher.kt
+++ b/src/main/kotlin/app/revanced/patcher/Patcher.kt
@@ -77,6 +77,7 @@ class Patcher(private val options: PatcherOptions) {
             this@integrations.callback = callback
         }
     }
+
     /**
      * Save the patched dex file.
      */
--- a/src/main/kotlin/app/revanced/patcher/Patcher.kt
+++ b/src/main/kotlin/app/revanced/patcher/Patcher.kt
@@ -78,6 +78,10 @@ class Patcher(private val options: PatcherOptions) {
         }
     }
 
+    fun clearMaps() {
+        MethodFingerprint.clearFingerprintResolutionLookupMaps()
+    }
+    
     /**
      * Save the patched dex file.
      */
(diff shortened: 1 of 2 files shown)


TheAabedKhan:
Yes. Since the maps needs to be cleared just for patching process for now, this is a good solution.

src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -77,7 +77,7 @@ class Patcher(private val options: PatcherOptions) {
             this@integrations.callback = callback
         }
     }
-
+
```
oSumAtrIX, line 80:
```suggestion
```

src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -392,4 +392,4 @@ class Patcher(private val options: PatcherOptions) {
  * @param patchInstance The instance of the [Patch] that was applied.
  * @param success The result of the [Patch].
  */
-internal data class ExecutedPatch(val patchInstance: Patch<Context>, val success: Boolean)
\ No newline at end of file
+internal data class ExecutedPatch(val patchInstance: Patch<Context>, val success: Boolean)
```
oSumAtrIX, line 395:
```suggestion
internal data class ExecutedPatch(val patchInstance: Patch<Context>, val success: Boolean)```

File: src/main/kotlin/app/revanced/patcher/Patcher.kt [hunk 3]
```diff
@@ -77,7 +77,6 @@ class Patcher(private val options: PatcherOptions) {
             this@integrations.callback = callback
         }
     }
```
oSumAtrIX, line 79:
```suggestion
    }

```

oSumAtrIX (approved):

PR #214, opened 2023-08-13
build: allow usage of `local` env to use local maven

Ushie:


build.gradle.kts
```diff
@@ -12,13 +12,17 @@ repositories {
...
         maven {
             url = uri("https://maven.pkg.github.com/revanced/$repo")
             credentials {
                 username = githubUsername
                 password = githubPassword
             }
         }
+    } else {
+        println("Fetching from mavenLocal, unset your `local` variable if you want to use GitHub packages")
+        mavenLocal()
```
oSumAtrIX, line 25: This if statement should be outside of the forEach call. Additionally why is it needed? You can simply include the localMaven() repo besides the GitHub Maven repo.

Ushie: It is necessary, there are only a few options and all the others introduce unwanted behavior

mavenLocal() above GitHub:
- prioritizes local over GH
GitHub over mavenLocal():
- Script fails once it reaches the username and password section
Checking whether username/password exists before:
- No way to select mavenLocal() over GitHub without removing the variables or modifying the script

oSumAtrIX: Will Gradle still fail when both GitHub and local maven are provided when Gradle can resolve the package in maven local? Because if it does not fail, it should work as expected if you put mavenLocal() above GitHub.

Ushie: Yes, it fails because the creds can't be null

oSumAtrIX: If you switch the order of getting the credentials, Gradle should handle it gracefully. This means change the order like this:

```diff
- username = project.findProperty("gpr.user") as? String ?: System.getenv("GITHUB_ACTOR")
+ username =  System.getenv("GITHUB_ACTOR") ?: project.findProperty("gpr.user")
```

Ushie: Turns out GitHub Packages doesn't even require authentication to read public packages, why is the credentials section even included?

oSumAtrIX: Is this still planned?

PR #272, opened 2024-02-11
feat: Read and write arbitrary files in APK files

oSumAtrIX:
# About

This PR allows reading and writing arbitrary files in an APK file. Additionally, it allows deleting files from APK files. A `RawResourcePatch` class has been added, which has access to `ResourceContext` but ReVanced Patcher will not decode APK resources. A regular `ResourcePatch` can read and write arbitrary files from an APK file unless they are decoded to `PatcherConfig.apkFiles`. On an attempt to get a file from `PatcherConfig.apkFiles` if the second parameter is true, it will read and write the raw resource file from the original APK to `PatcherConfig.apkFiles` if it does not exist. With this commit, many APIs have been deprecated as well, such as `DomFileEditor` and instead, a `Document` has been added.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
 org.gradle.parallel = true
 org.gradle.caching = true
-version = 19.2.1-dev.2
+version = 19.2.1-dev.1
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,3 @@
 org.gradle.parallel = true
 org.gradle.caching = true
-kotlin.code.style = official
-version = 19.2.1-dev.1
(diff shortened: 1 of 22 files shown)


gradle.properties
```diff
@@ -1,4 +1,3 @@
 org.gradle.parallel = true
 org.gradle.caching = true
-kotlin.code.style = official
-version = 19.2.1-dev.1
+version = 19.2.1-dev.2
```
oSumAtrIX, line 3:
```suggestion
version = 19.2.1-dev.1
```

Aunali321 (approved):

PR #293, opened 2024-05-02
Refactor the functional interfaces declaration

Tahgolov:
According to the specification, Kotlin has its own way of declaring functional interfaces. https://kotlinlang.org/spec/declarations.html#functional-interface-declaration


oSumAtrIX:
By now changes have been made which made this PR obsolete, thanks regardless

PR #306, opened 2024-08-29
docs: tipo resource using rawResource

Taknok:
The `resourcePatch` value was constructed from the `RawResourcePatch` which is a bit misleading. Also use the 1st letter uppercase convention from Kotlin.

### DIFF
--- a/docs/2_patches_intro.md
+++ b/docs/2_patches_intro.md
@@ -94,21 +94,21 @@ Example of patches:
 
 ```kt
 @Surpress("unused")
-val bytecodePatch = BytecodePatch {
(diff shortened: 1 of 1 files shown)


oSumAtrIX (changes requested):
The functions are currently in lowercase. Changing the docs to uppercase when the actual APIs are lowercase is impossible. Can you link the convention regarding DSL builder functions being uppercase?

Taknok:
Ho, those are functions, sorry. Thus they are corrects. I edit the PR.

oSumAtrIX (approved):

PR #312, opened 2024-10-14
fix: Make it work on Android 12 and lower by using existing APIs

LisoUseInAIKyrios:
Fixes patching failure when using Manger and Android 12 or lower.

All other patcher code seems to compile just fine.

Changing the project Android SDK to 8.0 correctly detects the issue.  I'm not sure if Android 8.0 can be set in a gradle config file or somewhere else, as changing the project in Android Studio does not change any project files.

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/patch/BytecodePatchContext.kt
+++ b/src/main/kotlin/app/revanced/patcher/patch/BytecodePatchContext.kt
@@ -71,7 +71,7 @@ class BytecodePatchContext internal constructor(private val config: PatcherConfi
             if (patch !is BytecodePatch) return@forEachRecursively
 
             patch.extension?.use { extensionStream ->
-                RawDexIO.readRawDexFile(extensionStream, 1024, null).classes.forEach { classDef ->
+                RawDexIO.readRawDexFile(extensionStream, 0, null).classes.forEach { classDef ->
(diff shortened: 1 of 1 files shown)


oSumAtrIX (approved):
Simplified the code as RawDexIO already implements a reader. Using the expectedSize argument, we can specify a buffer size. Extension DEX are usually about 1MB iirc so should we keep the size?

src/main/kotlin/app/revanced/patcher/patch/BytecodePatchContext.kt
```diff
@@ -62,17 +62,16 @@ class BytecodePatchContext internal constructor(private val config: PatcherConfi
...
 
         forEachRecursively { patch ->
-            if (patch is BytecodePatch && patch.extension != null) {
+            if (patch !is BytecodePatch) return@forEachRecursively
 
-                val extension = patch.extension.readAllBytes()
-
-                RawDexIO.readRawDexFile(extension, 0, null).classes.forEach { classDef ->
+            patch.extension?.use { extensionStream ->
+                RawDexIO.readRawDexFile(extensionStream, 1024, null).classes.forEach { classDef ->
```
oSumAtrIX, line 74:
```suggestion
                RawDexIO.readRawDexFile(extensionStream, 0, null).classes.forEach { classDef ->
```

PR #321, opened 2024-11-09
perf: Execute patches in parallel

oSumAtrIX:


oSumAtrIX:
@LisoUseInAIKyrios I am seing ConcurrentModificationExceptions in some places. I have tried to synchronize by locking them, but it seems something else is needed to be done. Perhaps you know what? The YouTube patches execute in around 2 seconds in parallel, compared to 17 seconds in sequence.

src/main/kotlin/app/revanced/patcher/patch/BytecodePatchContext.kt
```diff
@@ -108,9 +108,11 @@ class BytecodePatchContext internal constructor(private val config: PatcherConfi
      *
      * @return A proxy for the class.
      */
-    fun proxy(classDef: ClassDef) = classes.proxyPool.find {
-        it.immutableClass.type == classDef.type
-    } ?: ClassProxy(classDef).also { classes.proxyPool.add(it) }
+    fun proxy(classDef: ClassDef) = synchronized(classes) {
```
oSumAtrIX, line 111: Here are some attempts to solve concurrency issues. DexLib seems to be already designed to be concurrent in mind, so we just need to figure out synchronization in patcher.

oSumAtrIX:
> Patching in parallel means patches can be applied in different order each time patching is done.

No order guarantees were made except for the dependencies being executed first. Concerning the patch dev, a patch can be executed at any time. If issues arise when patches are executed in different orders when allowed to, they should be fixed.

> Using a CopyOnWriteArrayList should fix all concurrency issues with writing while iterating, but it has the cost of copying everything on every write. If classes are not added often it should be ok, but fixing the synchronization would probably be better.

I was thinking about that too. Classes are read often but not modified during patch execution time. Though, after execution the mutable classes replace the old ones. What do you mean by "fixing the synchronization"?

src/main/kotlin/app/revanced/patcher/patch/BytecodePatchContext.kt
```diff
@@ -98,8 +98,23 @@ class BytecodePatchContext internal constructor(private val config: PatcherConfi
...
-    fun classBy(predicate: (ClassDef) -> Boolean) =
-        classes.proxyPool.find { predicate(it.immutableClass) } ?: classes.find(predicate)?.let { proxy(it) }
+    fun classBy(predicate: (ClassDef) -> Boolean): ClassProxy? {
+        synchronized(classes.proxyPool) {
+            val proxy = classes.proxyPool.find { predicate(it.immutableClass) }
+            if (proxy != null) return proxy;
+        }
+
+        val classDef: ClassDef?
+        synchronized(classes) {
```
oSumAtrIX, line 108: Is synchronization needed here? Patches can also access the classes list, should they also synchronize?

oSumAtrIX:
I am not sure why, but I am getting a ConcurrentModificationException here, even though the classes list is synchronized:

image

src/main/kotlin/app/revanced/patcher/util/ProxyClassList.kt
```diff
@@ -2,28 +2,32 @@ package app.revanced.patcher.util
...
 
     /**
      * Replace all classes with their mutated versions.
      */
-    internal fun replaceClasses() =
-        proxyPool.removeIf { proxy ->
-            // If the proxy is unused, return false to keep it in the proxies list.
-            if (!proxy.resolved) return@removeIf false
+    internal fun replaceClasses() {
+        synchronized(proxyPool) {
```
oSumAtrIX, line 19: This does not seem necessary. The function is called after all patches have executed.

oSumAtrIX:
Some patches probabaly also need to explicitly synchronize https://github.com/ReVanced/revanced-patches/commit/ab32ae2f007d0d62b2577682d6140f4842d044e0

oSumAtrIX:
> It would require patches to once again declare what fingerprints it will resolve, but that can be avoided if in the fingerprint constructors they add themselves to a shared list and that list is then resolved in parallel all at once

It isn't worth breaking modular design by creating a shared list. Fingerprints match when the patch is used. If patches were to execute in parallel so would the fingerprints, but I checked and the fingerprints for YT match in 5s in total, so its not really worth

PR #322, opened 2024-11-10
perf: increase efficiency by using coroutines

Axelen123:
This PR contains ideas I got while looking at discussions and code related to the `perf/parallel-patch-execution` branch.

# Changes
- Fingerprint matching and class searches are now done in parallel
- IO operations are now performed on IO threads
- Changed the Patch API to use coroutines

# TODOs
- [ ] Fix resource patch data races
- [ ] Improve API
- [ ] Cleanup

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/Fingerprint.kt
+++ b/src/main/kotlin/app/revanced/patcher/Fingerprint.kt
@@ -84,9 +84,11 @@ class Fingerprint internal constructor(
         }
         if (match != null) return match
 
-        classes.forEach { classDef ->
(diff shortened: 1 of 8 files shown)


Ushie:
Maybe this PR should target the other branch instead of dev

Axelen123:
The patcher has a coroutine-aware lock to prevent execution of other bytecode patches until the current one has finished. Resource patches are still being executed on multiple threads, but that will be changed later. On my PC, patch execution time goes from 9,5 seconds to about 5,6. I have not tested it yet but phones will probably benefit more from this.

PR #331, opened 2025-01-05
fix: Support UTF-8 chars when compiling instructions in Smali in non UTF-8 environments

### LINKED ISSUE #330: bug: invalid smali strings  on OS where UTF-8 is not default encoding.
### Bug description

When I wanna use `const-string` instructions with strings such as `:§l§a#3` or **contained characters from another languages** it's compiled as invalid.
Here, you can see comparation between Linux(valid)/Windows(invalid):
Image

### Solution

I think this bug related wtih `Locale.ENGLISH` used in the [InlineSmaliCompiler](https://github.com/ReVanced/revanced-patcher/blob/main/src/main/kotlin/app/revanced/patcher/util/smali/InlineSmaliCompiler.kt)

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

Tahgolov:


src/main/kotlin/app/revanced/patcher/util/smali/InlineSmaliCompiler.kt
```diff
@@ -50,7 +50,7 @@ class InlineSmaliCompiler {
                     registers,
                     instructions,
                 )
-            val reader = InputStreamReader(input.byteInputStream())
+            val reader = InputStreamReader(input.byteInputStream(), Charsets.UTF_8)
```
oSumAtrIX, line 53: Why is UTF_8 chosen? Is it not arbitrary?

Tahgolov: Initially, calls InputStreamReader's constructor where is used default charset specific for OS. Next, `byteUnputStream()` uses UTF_8 always as default argument value. If the default charset is not UTF_8(ex. Windows) then after compiling smali code strings used in the instructions will be invalid.

oSumAtrIX: > See the attached issue. If fingerprint declare non ascii string constants they fail to match. Reading UTF8 fixes it. Or at least this is what I understand.

Yes I understood that much, however what if I want to use a charset that isn't UTF-8. Arbitrarily hardcoding an encoding may be a problem.

oSumAtrIX: I suppose, however a dynamic solution is probably better?

Tahgolov: The problem is conflict between 2 encodings when environment encoding is not utf-8. To resolve conflict we need to use only one encoding. I think we can use utf-8 by default since it is supported almost everywhere, but also we can use the environment encoding instead

oSumAtrIX (approved):

PR #332, opened 2025-01-19
fix: Add back missing log by naming logger correctly

kitadai31:
`this::javaClass.name` returns a wrong name

```kt
println(this::javaClass.name)
// javaClass

println(this::class.java.name)
// app.revanced.patcher.patch.BytecodePatchContext
```

And the logger is configured to output logs whose loggerName starts with "app.revanced".
As a result, log messages in `BytecodePatchContext` was not printed on both CLI and Manager.
This PR fixes this problem.


kitadai31:
Additional information: Other usage of `Logger.getLogger()` are fine
Only this part is using `javaClass` in `getLogger()`

oSumAtrIX:
I debugged this over half an hour and couldn't figure out what was wrong, lol

oSumAtrIX (approved):

PR #338, opened 2025-04-12
fix: Convert surrogate pairs into single unicode characters

LisoUseInAIKyrios:
Possible fix for surrogate character compilation error with ReVanced Manager.

Code cherry-picked from https://github.com/Taknok/revanced-patcher/commit/14e8e4be928f593619c76e3cca08225465bea34b


src/main/kotlin/app/revanced/patcher/util/Document.kt
```diff
@@ -23,6 +24,22 @@ class Document internal constructor(
         readerCount.merge(file, 1, Int::plus)
     }
 
+    private fun fixSurrogatePairs(xml: String): String {
+        val regex = """&#(\d+);&#(\d+);""".toRegex()
```
oSumAtrIX, line 28: This looks like a hacky attempt to fix the issue. Isn't it an encoding issue? Perhaps the correct encoding can be specified for the writer instead?

Issue #163, opened 2023-03-13
feat: abstract patch options


oSumAtrIX:
### Type

Functionality

### Issue

ReVanced Patcher should only provide primitive types for patch options such as an `IntOption` or `StringOption`. This creates complications when abstractions are required at a higher level such as when a patch has a color option. ReVanced Manager for example can not differentiate between colors and strings. 

### Feature

Allow options to be extendable. Patches can provide their own sets of abstractions for `PatchOption` such as a `ColorOption` inheriting from the primitive `StringOption`. These options should be in a separate module, published as a maven library to be used by higher-level applications such as ReVanced Manager. The patches would use the abstract options and shade them at compile time. Higher-level applications would use the library to compile against. ReVanced Manager could then use `ColorOption` to display a color picker. Upon loading the patches, `ClassLoader` will be aware of the abstract options at runtime. If the higher-level application is not aware of the abstract options, such as when not compiling against the abstract options library, then it will treat the options as the primitive option they inherit from. This is the instance when a patch, for example, does not make use of our abstract options but for example their own. ReVanced Manager would not be aware of these options and thus fall back to the primitive option unless ReVanced Manager is compiled against these abstract options. By default, ReVanced high-level applications like ReVanced Manager should compile against ReVanced abstract options as a base abstraction. Other patches outside of ReVanced are free to use the ReVanced library or extend the primitive options on their own, with the limitation, of high-level applications falling back to the primitive type.

TLDR:

Patches provide abstract options such as `ColorOption` by inheriting from ReVanced Patcher primitive options. High-level applications can be aware of those options by compiling against them. If they are unaware of the abstract options, they can fall back to the primitive type the option inherits from.

### Motivation

This change would come in favor of a higher level for UX. A color picker is more favorable than an input box with the necessity of the user having to infer the real type of the option such as when a text box is provided but a hex color is expected.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

PR #339, opened 2025-04-19
fix: Encode XML files as UTF-8 to fix compilation of resources

Taknok:
The android implementation of the transformer encodes the emoji in a form called "surrogate", but aapt and apktool do not understand the surrogate form. Thus, detect surrogate and recompute the traditional code point value before writing it to file allowing aapt and apktool to rebuild the app successfully.

It should fix some cases (mainly for google app) of https://github.com/ReVanced/revanced-manager/issues/2142

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/util/Document.kt
+++ b/src/main/kotlin/app/revanced/patcher/util/Document.kt
@@ -4,7 +4,9 @@ import org.w3c.dom.Document
 import java.io.Closeable
 import java.io.File
 import java.io.InputStream
+import java.io.StringWriter
 import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.transform.OutputKeys
 import javax.xml.transform.TransformerFactory
 import javax.xml.transform.dom.DOMSource
 import javax.xml.transform.stream.StreamResult
@@ -35,14 +37,19 @@ class Document internal constructor(
             }
 
             it.outputStream().buffered().use { stream ->
-                TransformerFactory.newInstance()
-                    .newTransformer()
-                    .transform(DOMSource(this), StreamResult(stream))
+                val transformer = TransformerFactory.newInstance().newTransformer()
+                // Set to UTF-16 but encode as UTF-8 to prevent surrogate pairs from being escaped to broken numeric character references.
+                if (isAndroid) {
+                    transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
+                    transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
+                }
+                transformer.transform(DOMSource(this), StreamResult(stream))
             }
         }
     }
 
     private companion object {
         private val readerCount = mutableMapOf<File, Int>()
+        private val isAndroid = System.getProperty("java.runtime.name").equals("Android Runtime")
     }
 }


Taknok:
I tried the modifications in the abandoned PR without success for the rebuild of the google phone app.

Should I rebase my branch on dev ?

oSumAtrIX:
I dislike band-aid solutions. For a couple of painful days I am experimenting a proper solution such as using a different transformer implementation like Saxon. Hopefully a proper solution can be found.

Taknok:
I agree, but the cleanest would be to fix the android transformer (to keep unicode point) or the aapt code (to handle surrogate form). I was not sure about introducing a new lib in the source code. Thus, I opted for the less impacting solution.

The wrapped solution or a argument for document() could be possible to. I could update the PR for this.

Taknok:
The core of the issue lies in the SDK based on OpenJDK, which in turn uses Apache Xalan. Java represents characters using UTF-16 ([source](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/String.java)). As a result, high Unicode code points are split into surrogate pairs (two char values). When parsed by Xalan's TreeWalker ([link](https://github.com/apache/xalan-j/blob/trunk/src/org/apache/xml/serializer/TreeWalker.java#L246)), these pairs are processed as two separate characters.

Subsequently, ToXMLStream or ToStream fails to recognize the two utf16 char as a single Unicode character and encodes each part separately, producing an incorrect surrogate representation. This is a known issue, tracked by the following tickets:
- https://issues.apache.org/jira/browse/XALANJ-2560
-https://issues.apache.org/jira/browse/XALANJ-2419

The ToHTMLStream has been fixed ([PR #163](https://github.com/apache/xalan-java/pull/163)), but no release has been published since 2023, so the fix is not available in official SDKs like Android or OpenJDK.

Conclusion: A fix for the Xalan transformer is unlikely to be available in the near future.

I also explored using the Saxon-HE transformer. Unfortunately, Saxon is not designed to support Android. During transformer creation, it depends on javax.xml.transform.stax and javax.xml.stream—two packages that Google chose not to include in Android’s core Java libraries. You can verify this in the list of supported packages: https://developer.android.com/reference/packages.html

To use Saxon, we would need to port those two packages to Android, which is likely beyond the scope of this repository.

That leaves us with three options:
1. Adopt a third-party XSLT transformer that supports Android. But which one ?
2. Use the regex-based approach proposed in this PR. On 10 runs I have an average of 1m21s for patching the google phone app
3. Convert the full string to UTF-16 and back to UTF-8, letting the cast merge surrogate bytes correctly On 10 runs I have an average of 1m18s for patching the google phone app

Regex based : https://github.com/Taknok/revanced-patcher/commit/c20e804b2b1a3672c1df79e96e2fb456b42de54a
UTF casting : https://github.com/Taknok/revanced-patcher/commit/ba57bcbaa13349e7ff1c1ddf74d4ff929438cbf8

@oSumAtrIX What would be the best ? the 3rd option with an optional parameter for the `context.document(resource, enforce_utf8=true).use...` ?

oSumAtrIX:
I was thinking of a more efficient solution which is to hook the output stream and modify it. This avoids writing to string, modifying it, then writing it to the file:


ChatGPT gave this as a starter:

```kt
import java.io.OutputStream
import java.io.Writer
import java.io.OutputStreamWriter

class UnicodeCorrectingWriter(outputStream: OutputStream, charset: String = "UTF-8") : Writer() {
    private val writer = OutputStreamWriter(outputStream, charset)
    private var pendingHighSurrogate: Char? = null

    override fun write(cbuf: CharArray, off: Int, len: Int) {
        for (i in off until off + len) {
            val ch = cbuf[i]
            if (Character.isHighSurrogate(ch)) {
                pendingHighSurrogate = ch
            } else if (Character.isLowSurrogate(ch) && pendingHighSurrogate != null) {
                val codePoint = Character.toCodePoint(pendingHighSurrogate!!, ch)
                writer.write(String(Character.toChars(codePoint)))
                pendingHighSurrogate = null
            } else {
                if (pendingHighSurrogate != null) {
                    // Write invalid pending high surrogate as-is
                    writer.write(pendingHighSurrogate!!.toInt())
                    pendingHighSurrogate = null
                }
                writer.write(ch.toInt())
            }
        }
    }

    override fun flush() {
        if (pendingHighSurrogate != null) {
            writer.write(pendingHighSurrogate!!.toInt())
            pendingHighSurrogate = null
        }
        writer.flush()
    }

    override fun close() {
        flush()
        writer.close()
    }
}
```

I haven't tested it but maybe this works. Would merge if this approach can be used.

Taknok:
I will look into it, thank for the hint

Taknok:
After quick look, the code never enter the highsurrogate, thus it never correctly encode the emoji.
```
if (Character.isHighSurrogate(ch)) {
  pendingHighSurrogate = ch
```
Furthermore it takes more than 4min to reach the crash  when repacking the res. I will continue to look into it to make it works, but it seems to not be hethe best performance strategy.

Taknok:
So unfortunately, the chatGPT method does not work as is, as the ToStream serializer write char by char to the stream when we reach this block : https://android.googlesource.com/platform/external/apache-xml/+/650a6cfd4d6b2d38b88ada03694ae19cc448d07b/src/main/java/org/apache/xml/serializer/ToStream.java#1609
The writer has to record the state of the potential surrogate and process the logic (as done in the xalan updated ToStream.java code).
The performance of this method is the lower among all. It tooks 1m42s average (when ~1m20s for the others). Due to the complexity of the writer, this method seems to be the less futur proof as it depends of the calls of ToStream.

My tracking branch: https://github.com/Taknok/revanced-patcher/commit/ca2de8eee31b284ec0ca282ee57a35dd660fe7fd

My opinion one the techniques:
1. UTF casting https://github.com/Taknok/revanced-patcher/commit/ba57bcbaa13349e7ff1c1ddf74d4ff929438cbf8
2. regex https://github.com/Taknok/revanced-patcher/commit/c20e804b2b1a3672c1df79e96e2fb456b42de54a
3. writer https://github.com/Taknok/revanced-patcher/commit/ca2de8eee31b284ec0ca282ee57a35dd660fe7fd

Available on discord if needed.

oSumAtrIX:
If this is all and a conditional check is added, I'd merge it with a comment referencing the Android issue tracker

Taknok:
There is not surrogate issue on JRE as the openjdk correctly implement the conversion. Thus the conditional check is still needed ?

I will try without the UTF8 , but I think y my tests it was not working without

Taknok:
Here are the requested modification. Tell me if you want some adjustments.

kitadai31:
It worked.
I patched GPhotos on Manager successfully with Patches 5.19.0 (the last patches version that photos compiling is not fixed).

Taknok:
You can also use google phone app with this patch to check if issue is fixed : https://github.com/Taknok/revanced-patches

src/main/kotlin/app/revanced/patcher/util/Document.kt
```diff
@@ -13,6 +14,12 @@ class Document internal constructor(
     inputStream: InputStream,
 ) : Document by DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(inputStream), Closeable {
     private var file: File? = null
+    private val isAndroid = try {
+        Class.forName("android.os.Build")
+        true
+    } catch (e: ClassNotFoundException) {
+        false
+    }
```
kitadai31, line 22: Move this to companion object so that you can avoid checking this every time

Also, `System.getProperty("java.runtime.name") == "Android Runtime"` is simpler

(This property has probably not been changed since the beginning, and even Android 4.x, before the introduction of ART, it was “Android Runtime”.)

src/main/kotlin/app/revanced/patcher/util/Document.kt
```diff
@@ -35,9 +42,11 @@ class Document internal constructor(
...
             it.outputStream().use { stream ->
-                TransformerFactory.newInstance()
-                    .newTransformer()
-                    .transform(DOMSource(this), StreamResult(stream))
+                val transformer = TransformerFactory.newInstance().newTransformer()
+                if (isAndroid) {
+                    transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
+                }
+                transformer.transform(DOMSource(this), StreamResult(stream))
             }
```
kitadai31, line 50: Not directly related to this PR, but I think outputStream is not necessary here.
`StreamResult(it)` will be fine

This appears to be a leftover of previous versions of Patcher, which took OutputStream as a constructor argument of DomFileEditor 👀

kitadai31:
Nope, outputting in UTF-16 caused problems when patching other apps
Converting to UTF-8 is probably necessary

Sorry for jumping to conclusions...

```
- Device Info
ReVanced Manager: 1.25.0-dev.1
Model: Pixel 3
Android version: 15
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: Yes

- Patch Info
App: com.google.android.youtube v20.12.46 (Suggested: 20.12.46)
Patches version: v5.22.0
Patches added: Default
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: false
Version compatibility check: true
Show universal patches: false
Patches source: revanced/revanced-patches

- Logs
Reading APK
Decoding app manifest
Loading patches
Deleting existing temporary files directory
Decoding resources
Initializing lookup maps
Executing patches
Applied 52 patches
GmsCore support failed: app.revanced.patcher.patch.PatchException: The patch "GmsCore support" depends on "ResourcePatch", which raised an exception:
app.revanced.patcher.patch.PatchException: Unexpected token (position:TEXT ï¿½ï��@1:8 in java.io.InputStreamReader@aec901d) 
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend$execute(Unknown Source:143)
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend$execute(Unknown Source:65)
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:190)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:2)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:1)
	at kotlinx.coroutines.flow.SafeFlow.collect(Unknown Source:102)
	at app.revanced.manager.flutter.MainActivity$runPatcher$1$patcherResult$1$1.invokeSuspend(Unknown Source:74)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:11)
	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:134)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(Unknown Source:91)
	at kotlinx.coroutines.JobKt.runBlocking$default(Unknown Source:104)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$34(Unknown Source:360)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$_NYn7P0Ss3RCL0-SqEFL-opMsQw(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda6.run(Unknown Source:18)
	at java.lang.Thread.run(Thread.java:1119)
Caused by: org.xml.sax.SAXParseException: Unexpected token (position:TEXT ï¿½ï��@1:8 in java.io.InputStreamReader@aec901d) 
	at org.apache.harmony.xml.parsers.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:147)
	at javax.xml.parsers.DocumentBuilder.parse(DocumentBuilder.java:107)
	at app.revanced.patcher.util.Document.<init>(SourceFile:2)
	at app.revanced.patcher.util.Document.<init>(SourceFile:4)
	at app.revanced.patcher.patch.ResourcePatchContext.document(SourceFile:2)
	at app.revanced.patches.shared.misc.gms.GmsCoreSupportPatchKt.gmsCoreSupportResourcePatch$lambda$35$lambda$34$addSpoofingMetadata(GmsCoreSupportPatch.kt:568)
	at app.revanced.patches.shared.misc.gms.GmsCoreSupportPatchKt.gmsCoreSupportResourcePatch$lambda$35$lambda$34(GmsCoreSupportPatch.kt:622)
	at app.revanced.patches.shared.misc.gms.GmsCoreSupportPatchKt.$r8$lambda$lRRewJm6n_a-02ysO3PqlF2CFs8(Unknown Source:0)
	at app.revanced.patches.shared.misc.gms.GmsCoreSupportPatchKt$$ExternalSyntheticLambda3.invoke(Unknown Source:15)
	at app.revanced.patcher.patch.Patch.execute(Unknown Source:7)
	at app.revanced.patcher.patch.ResourcePatch.execute$revanced_patcher(Unknown Source:9)
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend$execute(Unknown Source:124)
	... 14 more

	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend$execute(Unknown Source:113)
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:190)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:2)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:1)
	at kotlinx.coroutines.flow.SafeFlow.collect(Unknown Source:102)
	at app.revanced.manager.flutter.MainActivity$runPatcher$1$patcherResult$1$1.invokeSuspend(Unknown Source:74)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:11)
	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:134)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(Unknown Source:91)
	at kotlinx.coroutines.JobKt.runBlocking$default(Unknown Source:104)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$34(Unknown Source:360)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$_NYn7P0Ss3RCL0-SqEFL-opMsQw(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda6.run(Unknown Source:18)
	at java.lang.Thread.run(Thread.java:1119)
Change package name failed: app.revanced.patcher.patch.PatchException: The patch "Change package name" raised an exception: app.revanced.patcher.patch.PatchException: Unexpected token (position:TEXT ï¿½ï��@1:8 in java.io.InputStreamReader@9750f0a) 
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:320)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:2)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:1)
	at kotlinx.coroutines.flow.SafeFlow.collect(Unknown Source:102)
	at app.revanced.manager.flutter.MainActivity$runPatcher$1$patcherResult$1$1.invokeSuspend(Unknown Source:74)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:11)
	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:134)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(Unknown Source:91)
	at kotlinx.coroutines.JobKt.runBlocking$default(Unknown Source:104)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$34(Unknown Source:360)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$_NYn7P0Ss3RCL0-SqEFL-opMsQw(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda6.run(Unknown Source:18)
	at java.lang.Thread.run(Thread.java:1119)
Caused by: org.xml.sax.SAXParseException: Unexpected token (position:TEXT ï¿½ï��@1:8 in java.io.InputStreamReader@9750f0a) 
	at org.apache.harmony.xml.parsers.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:147)
	at javax.xml.parsers.DocumentBuilder.parse(DocumentBuilder.java:107)
	at app.revanced.patcher.util.Document.<init>(SourceFile:2)
	at app.revanced.patcher.util.Document.<init>(SourceFile:4)
	at app.revanced.patcher.patch.ResourcePatchContext.document(SourceFile:2)
	at app.revanced.patches.all.misc.packagename.ChangePackageNamePatchKt.changePackageNamePatch$lambda$9$lambda$8(ChangePackageNamePatch.kt:78)
	at app.revanced.patches.all.misc.packagename.ChangePackageNamePatchKt.$r8$lambda$6AkdPHArk5z26dp2iCI31m62ogY(Unknown Source:0)
	at app.revanced.patches.all.misc.packagename.ChangePackageNamePatchKt$$ExternalSyntheticLambda4.invoke(Unknown Source:6)
	at app.revanced.patcher.patch.Patch.finalize(Unknown Source:9)
	at app.revanced.patcher.patch.ResourcePatch.finalize$revanced_patcher(Unknown Source:9)
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:308)
	... 12 more

	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:382)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:2)
	at app.revanced.patcher.Patcher$invoke$1.invoke(SourceFile:1)
	at kotlinx.coroutines.flow.SafeFlow.collect(Unknown Source:102)
	at app.revanced.manager.flutter.MainActivity$runPatcher$1$patcherResult$1$1.invokeSuspend(Unknown Source:74)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:11)
	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:134)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(Unknown Source:91)
	at kotlinx.coroutines.JobKt.runBlocking$default(Unknown Source:104)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$34(Unknown Source:360)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$_NYn7P0Ss3RCL0-SqEFL-opMsQw(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda6.run(Unknown Source:18)
	at java.lang.Thread.run(Thread.java:1119)
Caused by: app.revanced.patcher.patch.PatchException: Unexpected token (position:TEXT ï¿½ï��@1:8 in java.io.InputStreamReader@9750f0a) 
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:320)
	... 12 more
Caused by: org.xml.sax.SAXParseException: Unexpected token (position:TEXT ï¿½ï��@1:8 in java.io.InputStreamReader@9750f0a) 
	at org.apache.harmony.xml.parsers.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:147)
	at javax.xml.parsers.DocumentBuilder.parse(DocumentBuilder.java:107)
	at app.revanced.patcher.util.Document.<init>(SourceFile:2)
	at app.revanced.patcher.util.Document.<init>(SourceFile:4)
	at app.revanced.patcher.patch.ResourcePatchContext.document(SourceFile:2)
	at app.revanced.patches.all.misc.packagename.ChangePackageNamePatchKt.changePackageNamePatch$lambda$9$lambda$8(ChangePackageNamePatch.kt:78)
	at app.revanced.patches.all.misc.packagename.ChangePackageNamePatchKt.$r8$lambda$6AkdPHArk5z26dp2iCI31m62ogY(Unknown Source:0)
	at app.revanced.patches.all.misc.packagename.ChangePackageNamePatchKt$$ExternalSyntheticLambda4.invoke(Unknown Source:6)
	at app.revanced.patcher.patch.Patch.finalize(Unknown Source:9)
	at app.revanced.patcher.patch.ResourcePatch.finalize$revanced_patcher(Unknown Source:9)
	at app.revanced.patcher.Patcher$invoke$1.invokeSuspend(Unknown Source:308)
	... 12 more
Compiling modified resources
/data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/apk/AndroidManifest.xml:1: error: not well-formed (invalid token).
An error occurred:
brut.androlib.exceptions.AndrolibException: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~uxzK18kQmnPAJcAbOvR0jw==/app.revanced.manager.flutter.development-yPDJNdvuT9xMrXHjOgBweQ==/lib/arm64/libaapt2.so, link, -o, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/patched/resources/resources.apk, --package-id, 127, --min-sdk-version, 26, --target-sdk-version, 35, --version-code, 1553610176, --version-name, 20.12.46, --no-auto-version, --no-version-vectors, --no-version-transitions, --no-resource-deduping, --allow-reserved-package-id, --warn-manifest-validation, -e, /data/user/0/app.revanced.manager.flutter.development/cache/APKTOOL9182088485613914028.tmp, -0, arsc, -I, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/1.apk, --manifest, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/apk/AndroidManifest.xml, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/apk/build/resources.zip]
	at app.revanced.patcher.patch.ResourcePatchContext.get(SourceFile:146)
	at app.revanced.patcher.Patcher.get(Unknown Source:18)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$34(Unknown Source:391)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$_NYn7P0Ss3RCL0-SqEFL-opMsQw(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda6.run(Unknown Source:18)
	at java.lang.Thread.run(Thread.java:1119)
Caused by: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~uxzK18kQmnPAJcAbOvR0jw==/app.revanced.manager.flutter.development-yPDJNdvuT9xMrXHjOgBweQ==/lib/arm64/libaapt2.so, link, -o, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/patched/resources/resources.apk, --package-id, 127, --min-sdk-version, 26, --target-sdk-version, 35, --version-code, 1553610176, --version-name, 20.12.46, --no-auto-version, --no-version-vectors, --no-version-transitions, --no-resource-deduping, --allow-reserved-package-id, --warn-manifest-validation, -e, /data/user/0/app.revanced.manager.flutter.development/cache/APKTOOL9182088485613914028.tmp, -0, arsc, -I, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/1.apk, --manifest, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/apk/AndroidManifest.xml, /data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/revanced-temporary-files/apk/build/resources.zip]
	at brut.util.OS.exec(Unknown Source:72)
	at app.revanced.patcher.patch.ResourcePatchContext.get(SourceFile:142)
	... 5 more
Something went wrong:
PathNotFoundException: Cannot copy file to '/data/user/0/app.revanced.manager.flutter.development/files/lastPatchedApp.apk', path = '/data/user/0/app.revanced.manager.flutter.development/files/patcher/tmp-KGGWPD/out.apk' (OS Error: No such file or directory, errno = 2)
```

Taknok:
Draft waiting UTF8 revert

src/main/kotlin/app/revanced/patcher/util/Document.kt
```diff
@@ -34,15 +35,18 @@ class Document internal constructor(
...
-                TransformerFactory.newInstance()
-                    .newTransformer()
-                    .transform(DOMSource(this), StreamResult(stream))
+            it.outputStream().use {
+                val transformer = TransformerFactory.newInstance().newTransformer()
+                if (isAndroid) {
+                    transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
+                }
+                transformer.transform(DOMSource(this), StreamResult(it))
             }
```
kitadai31, line 44: No, I meant removing `it.outputStream().use { }` and passing the `File` to `StreamResult()` directly
My bad

Taknok:
```kotlin
val transformer = TransformerFactory.newInstance().newTransformer()
if (isAndroid) {
    transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
}
transformer.transform(DOMSource(this), StreamResult(it))
```
?

Taknok: We have to use the outputStream() to be able to call write(). As we use outoutStream() we have to use "use" to ensure the descriptor is closed

kitadai31:
In the end, we went back to the approach mentioned in LisoUseInAIKyrios's first comment.

But I'm thinking of using a FileWriter directly instead of using a StringWriter.
Since FileWriter uses the system default encoding, and Android's default encoding is always UTF-8, I expect it to write in UTF-8.

If this were possible, the serialized result would be written directly to a file, so there would be no performance impact.
I'll try this tomorrow

Taknok:
This function compile youtube and google phone without error:

```kotlin
    override fun close() {
        file?.let {
            if (readerCount[it]!! > 1) {
                throw IllegalStateException(
                    "Two or more instances are currently reading $it." +
                        "To be able to close this instance, no other instances may be reading $it at the same time.",
                )
            } else {
                readerCount.remove(it)
            }

            val transformer = TransformerFactory.newInstance().newTransformer().apply {
                if (isAndroid) {
                    setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
                    setOutputProperty(OutputKeys.ENCODING, "UTF-16")
                }
            }

            it.writer().use { writer ->
                transformer.transform(DOMSource(this), StreamResult(writer))
            }
        }
    }
```

oSumAtrIX:
Why is 

```
                    setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
```

needed?

Taknok:
Because we want to have an UTF-8 XML at the end, thus we have to explicitly say to not write the encoding as we perform the conversion : it.writer() set by default a charsets.UTF-8

oSumAtrIX:
So as I understand the transformer writes as utf16 and then the writer as utf8? Why are we omitting the xml declaration. It doesn't have to do anything with encoding from what I am seeing

Taknok:
The origin:
1. App
2. Apktool decode XML to UTF-8
3. Java read UTF-8 and set it well in memory, string IB memory handles wells UTF-8. When writing to a file on the string it calls .toCharsArray() which encode it into a 16bits chars list (but some unicode needs overflow and are split into 2 16bits chars)
4.1 openjdk, when writing, is aware of this potential issue. So it handle the writing by checking if the 16chars is a high point unicode and if the encoding is UTF-8, if yes, it goes to a special flow writing the 16charsunicode.
4.2 Android is not aware of it, it thinks that it is handling UTF-8 that are only 16chars, if this unicode is unknown, so it writes it as an entity as would do JavaScript.
5. Apktool recompiles the res, it do not understand surrogate unicode, if it reads one (like in Android senario) it crashes.

The fix:
1. In memory we have a UTF-8 string with correct high point emoji.
2. The toStream class will be called and split the strings with toCharsArray() (else we need to rewrite the whole ToStream from xalan, I tried this but there is private class etc in xalan increasing complexity and code copy).
3. We tell to the writer that it will handle 16 bits char (UTF-16), but the charset used will be UTF-8 (default value of the file writers), so the output file should not have an xml header forcing to UTF-18 as we used UTF-8 charset.
4. Aapt2 reads the file with UTF-8 and high point encoded value.


If you want to follow the flow in the debugging mode, you have to set a breaking in the characters() function for the ToStream.java class (called by the xml TreeWalker.java) and follow step by step the code for each case

Taknok:
> So as I understand the transformer writes as utf16 and then the writer as utf8?

Yes.

We are omitting because the resulting xml file is written with UTF-8 and aapt2 will read using UTF-8. Setting the headers in the xml file may leads to an aapt2 reading UTF-8 value thinking it is utf-16 (as declared in the headers)

Taknok:
I know this is hacky, but it google was consistent and keep only 16bits chars with surrogate support OR UTF-8 correct writing this would not be needed :/

oSumAtrIX:
Can't you set the encoding in the header to utf8 explicitly but write in utf16?

Taknok:
I did not find a way to do it, but I also did not find a proof that is not feasible.

kitadai31:
For JRE, there is no need to use Writer
I suggest this code
Also I tried adding some comments, I hope it helps

```kt
            val transformer = TransformerFactory.newInstance().newTransformer()
            if (isAndroid) {
                // Set to UTF-16 to prevent surrogate pairs from being escaped to broken numeric character references.
                transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
                // The XML declaration will have encoding="UTF-16", but we're going write it back in UTF-8, so omit it.
                transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
                // Use FileWriter to output the XML file in UTF-8 encoding.
                it.writer().use { writer ->
                    transformer.transform(DOMSource(this), StreamResult(writer))
                }
            } else {
                transformer.transform(DOMSource(this), StreamResult(it))
            }
```

src/main/kotlin/app/revanced/patcher/util/Document.kt
```diff
@@ -34,15 +36,24 @@ class Document internal constructor(
...
-                    .newTransformer()
-                    .transform(DOMSource(this), StreamResult(stream))
+            val transformer = TransformerFactory.newInstance().newTransformer()
+            if (isAndroid) {
+                // Set to UTF-16 to prevent surrogate pairs from being escaped to broken numeric character references.
+                transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
+                // The XML declaration will have encoding="UTF-16", but we're going to write it back in UTF-8, so omit it.
+                transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
+                // Use FileWriter to output the XML file in UTF-8 encoding.
+                it.writer().use { writer ->
```
kitadai31, line 46:
```suggestion
                it.bufferedWriter().use { writer ->
```

I forgot about buffering
Just applying this suggestion is OK

Taknok: True, buffered writer is useful when doing many writes, and ToStream write char by char, thus it will improve perf.

Taknok:
Perf test :
Manager main branch + patcher main with surrogate fix : 334s
Manager main branch + patcher main branch vanilla : 349s

Taknok:
The (slightly) better performance is due to the buffered writer that buffer the chars wrote one by one. A main branch build with a buffered writer (https://github.com/Taknok/revanced-patcher/tree/main-buffered) have a patch time (for youtube with default selection) of 290s

Taknok:
Hi @kitadai31 
This is a gentle reminder about this PR I opened a month ago, no rush at all. Whenever you have a chance to take a look, I'd really appreciate your feedback.

Taknok:
Euh... ok but the way of how is formulated your message seems that the fix of the surrogate crash is not desirable ? O.o

kitadai31:
I think this PR is ready

src/main/kotlin/app/revanced/patcher/util/Document.kt
```diff
@@ -34,15 +36,24 @@ class Document internal constructor(
...
+            val transformer = TransformerFactory.newInstance().newTransformer()
+            if (isAndroid) {
+                // Set to UTF-16 to prevent surrogate pairs from being escaped to broken numeric character references.
+                transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
+                // The XML declaration will have encoding="UTF-16", but we're going to write it back in UTF-8, so omit it.
+                transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
+                // Use FileWriter to output the XML file in UTF-8 encoding.
+                it.bufferedWriter().use { writer ->
+                    transformer.transform(DOMSource(this), StreamResult(writer))
+                }
```
oSumAtrIX, line 48:
```suggestion
            // Set to UTF-16 but encode as UTF-8 to prevent surrogate pairs from being escaped to broken numeric character references.
            if (isAndroid) {
                transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16")
                transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes")
                it.bufferedWriter().use { writer ->
                    transformer.transform(DOMSource(this), StreamResult(writer))
                }
```

oSumAtrIX (approved):
If the comment is right, this is fine to merge

Taknok:
I will rebase this PR on the new dev

kitadai31:
@Taknok Hey, your last push broke everything!
A use of Writer is missing.
When UTF-16 is set to the output encoding, it must be output with Writer to make the final output file UTF-8.
So a Writer must be used in `isAndroid` block.

Also, after some testing, I found that OutputStream is same with or without `.buffered()`.
I created a custom `FileOutputStream` and `FileWriter` that record the number of writes, and measured the number of times each write was called, with and without wrapping it in Buffered, when output an XML file.

```
The number of times write() was called while outputting YT's public.xml.
MyFileOutputStream without buffering: 119
MyFileOutputStream with buffering: 119
MyFileWriter without buffering 1461632
MyFileWriter with buffering: 238
```

This result shows that when constructing a StreamResult, a Writer must be wrapped in Buffered, but an OutputStream does not need to be wrapped.
The XML Transformer is likely doing its own buffering if the target is an OutputStream.
Therefore, #347 (`perf: Use a buffered writer to reduce IO overhead`) was actually not necessary.

The source code I used to test:
https://gist.github.com/kitadai31/f087562d19a740e43ff472a8afcf623b

This PR basically needs to be reverted to the point of [dfb7f05](https://github.com/ReVanced/revanced-patcher/blob/dfb7f05da4cb0394dd1536b56ca0d74b97c9fe08/src/main/kotlin/app/revanced/patcher/util/Document.kt) (Before force-pushed)
I will open a PR to fix these.

Taknok:
Ok tag me in the PR you open

Taknok:
Thank you

PR #343, opened 2025-05-15
fix: Replace dex input streams with file references

hoo-dles:
This should fix the issue where the CLI cannot purge the temporary patch files because there's still file handles from hanging input streams. 

I originally was going to just make `PatcherResult` implement `Closeable`, but I think it makes more sense to let Revanced library open and close the streams instead of passing them around.

https://github.com/ReVanced/revanced-cli/issues/327

### DIFF
--- a/api/revanced-patcher.api
+++ b/api/revanced-patcher.api
@@ -171,9 +171,12 @@ public final class app/revanced/patcher/patch/BytecodePatchContext : app/revance
 public final class app/revanced/patcher/patch/Option {
 	public fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;)V
 	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;)V
+	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getDefault ()Ljava/lang/Object;
 	public final fun getDescription ()Ljava/lang/String;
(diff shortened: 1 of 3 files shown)


api/revanced-patcher.api
```diff
@@ -171,9 +171,12 @@ public final class app/revanced/patcher/patch/BytecodePatchContext : app/revance
 public final class app/revanced/patcher/patch/Option {
 	public fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;)V
 	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;)V
+	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;Ljava/lang/String;ZLkotlin/reflect/KType;Lkotlin/jvm/functions/Function2;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
hoo-dles, line 175: Not sure where these changes came from. Maybe previous merge that didn't run apiDump?

src/main/kotlin/app/revanced/patcher/patch/BytecodePatchContext.kt
```diff
@@ -152,7 +152,7 @@ class BytecodePatchContext internal constructor(private val config: PatcherConfi
                     DexIO.DEFAULT_MAX_DEX_POOL_SIZE,
                 ) { _, entryName, _ -> logger.info { "Compiled $entryName" } }
             }.listFiles(FileFilter { it.isFile })!!.map {
-                PatcherResult.PatchedDexFile(it.name, it.inputStream())
```
oSumAtrIX, line 155: The input stream can be simply closed here via a use block. No need to convert to a file.

hoo-dles: Just adding a `use` block here will leave scope and close the stream before `PatcherResult.applyTo()` is called.

This is why I advocate passing around a file reference instead of an open stream; it's much easier to reason about.

oSumAtrIX: The fact that Patcher is dumping the dex to a file is not something that any other library should take advantage of. It is merely done for caching purposes and no other purpose. In the most basic form patcher has no "requirements" for or does not require anyone to work with files. A stream is all that's needed. If a consuming library needs to convert it to a file it's free to, if one doesn't then there is no reason to require working with files in the first place. If the stream needs to be open, then the use block should not be used here and whatever consumes the stream should simply close the stream.

PR #347, opened 2025-06-19
perf: Use a buffered writer to reduce IO overhead

Taknok:
As ToSteam (used by XMLTransformer) writes char by char, using a buffered writer improves significantly the performance of the writing.

See #339 for the tests and the discussion


Taknok:
```kotlin
it.outputStream().buffered().use { output ->
    TransformerFactory.newInstance()
        .newTransformer()
        .transform(DOMSource(this), StreamResult(output))
}
```
Will force a buffered path, but it does not differ from the approach
```
it.bufferedWriter().use { writer ->
    TransformerFactory.newInstance()
        .newTransformer()
        .transform(DOMSource(this), StreamResult(writer))
}
```
FileOutputStream is not buffered by default, so when you write to a FileOutputStream, every write() call can result in a system-level I/O operation, which is what it was asked to fix/avoid in this PR no ?

oSumAtrIX (approved):

Issue #359, opened 2025-07-18
bug: Dexless APKs cause a `EmptyMultiDexContainerException`


Delta-official:
### Bug description

Currently, trying to patch an APK with no .dex files result in a `EmptyMultiDexContainerException` due to `BytecodeContext` always being loaded, regardless of whether the APK contains .dex files. For example, this can be encountered when patching the part of a split APK that contains native libraries.

### Error logs

```shell
INFO: Loading patches
lanchon.multidexlib2.EmptyMultiDexContainerException: split_config.arm64_v8a.apk
        at lanchon.multidexlib2.MultiDexIO.readMultiDexContainer(MultiDexIO.java:41)
        at lanchon.multidexlib2.MultiDexIO.readDexFile(MultiDexIO.java:33)
        at app.revanced.patcher.patch.BytecodePatchContext.<init>(BytecodePatchContext.kt:48)
        at app.revanced.patcher.PatcherContext.<init>(PatcherContext.kt:40)
        at app.revanced.patcher.Patcher.<init>(Patcher.kt:19)
        at app.revanced.cli.command.PatchCommand.run(PatchCommand.kt:288)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2030)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2465)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2457)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2419)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
        at picocli.CommandLine$RunLast.execute(CommandLine.java:2421)
        at picocli.CommandLine.execute(CommandLine.java:2174)
        at app.revanced.cli.command.MainCommandKt.main(MainCommand.kt:12)
```

### Solution

One solution would be to wait for https://github.com/ReVanced/revanced-patcher/issues/73. However, I'd like to make a PR that would catch `EmptyMultiDexContainerException` in `PatcherContext` and set a field that would disable bytecode patching behaviour. This could mitigate the issue and allow resources patches to run normally without being blocked.

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

[...]

PR #360, opened 2025-07-19
fix: Allow patcher to run on dex-less APKs

Delta-official:
Resolves https://github.com/ReVanced/revanced-patcher/issues/359

### DIFF
--- a/src/main/kotlin/app/revanced/patcher/Patcher.kt
+++ b/src/main/kotlin/app/revanced/patcher/Patcher.kt
@@ -28,6 +28,9 @@ class Patcher(private val config: PatcherConfig) : Closeable {
      * @param patches The patches to add.
      */
     operator fun plusAssign(patches: Set<Patch<*>>) {
+        // Filter out bytecode patches if bytecodeContext is null
+        val patches = patches.filterNotTo(mutableSetOf()) { (context.bytecodeContext == null && it is BytecodePatch) }
+
         // Add all patches to the executablePatches set.
         context.executablePatches += patches
 
@@ -99,7 +102,7 @@ class Patcher(private val config: PatcherConfig) : Closeable {
         logger.info("Initializing lookup maps")
 
         // Accessing the lazy lookup maps to initialize them.
-        context.bytecodeContext.lookupMaps
+        context.bytecodeContext?.lookupMaps
 
         logger.info("Executing patches")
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/app/revanced/patcher/Patcher.kt
```diff
@@ -28,6 +28,9 @@ class Patcher(private val config: PatcherConfig) : Closeable {
      * @param patches The patches to add.
      */
     operator fun plusAssign(patches: Set<Patch<*>>) {
+        // Filter out bytecode patches if bytecodeContext is null
```
oSumAtrIX, line 31: This shouldnt be done. What if I want to add and operate on dex files using bytecode patches

oSumAtrIX (changes requested):
Couple changes needed

src/main/kotlin/app/revanced/patcher/PatcherContext.kt
```diff
@@ -37,7 +41,12 @@ class PatcherContext internal constructor(config: PatcherConfig): Closeable {
     /**
      * The context for patches containing the current state of the bytecode.
      */
-    internal val bytecodeContext = BytecodePatchContext(config)
+    internal val bytecodeContext : BytecodePatchContext? = try {
```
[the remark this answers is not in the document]
Delta-official: > Instead of controlling flow with exception handling you should check if the APK has a dex. 

Why? This is a good way to detect a dexless APK that wouldn't involve practically repeating the logic of androlib.

> Right now you don't handle the third case and I don't know if this needs to be handled, or how.

The third case doesn't make sense. If you have a dex file, it *must* at least have the app entrypoint

src/main/kotlin/app/revanced/patcher/Patcher.kt
oSumAtrIX: What I mean is add a new class/dex to a dex less APK

src/main/kotlin/app/revanced/patcher/PatcherContext.kt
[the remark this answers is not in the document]
oSumAtrIX: I think the other comment motivates this one. What if I want to add classes to dexless APKs? I need a bytecodepatch context, i also would be able to empty the list of classes or add classes, so the three cases apply.

src/main/kotlin/app/revanced/patcher/Patcher.kt
Delta-official: I see what you mean now. I'm currently a little busy but I'll try to get address these sometime soon

oSumAtrIX:
So right now i am looking into multidexlib to be able to read dexless dex container, it will simply return an empty array, so the patcher doesnt have to account for an exception

oSumAtrIX:
Essentially this means, that the arrays/sets will be empty, on writeback, we can then check if there are any classes to write and write them back

Ushie:
This appears to be superseded by https://github.com/ReVanced/revanced-patcher/commit/d166863c95757203e5611f6051fded3e5a921c8c

Should this be closed? @oSumAtrIX

oSumAtrIX:
Yep, thanks

PR #378, opened 2025-10-16
feat: Add callback for patch loading exceptions

oSumAtrIX:
I am not sure why the constructor was changed in the API. It is supposed to be internal/private. The Dex/Jar constructors were meant to be private; it should be okay to "fix" without a breaking change.


src/main/kotlin/app/revanced/patcher/patch/Patch.kt
```diff
@@ -640,19 +656,24 @@ sealed class PatchLoader private constructor(
...
-                    loadClass(it)
-                }.flatMap {
-                    it.patchFields + it.patchMethods
-                }.filter {
-                    it.name != null
-                }.toSet()
-            }
+        private fun ClassLoader.loadPatches(
+            binaryClassNamesByPatchesFile: Map<File, Set<String>>,
+            onLoadPatchesException: (message: String, cause: Throwable) -> Unit
```
Axelen123: It would be good to also give the file as well so we can associate the error with the correct bundle in the UI.

Axelen123 (changes requested):

PR #329, opened 2025-01-01
feat: Match fingerprints by instruction filters

LisoUseInAIKyrios:
Adds instruction filters to support more flexible instruction fingerprinting.
&nbsp;

#### Fingerprints can still use opcode patterns, or they can use instruction filters that allow more precise matching.

Basic support exists for matching instructions using:
- MethodFilter: Specifies defining class, method name, parameters, return type, etc.  (Sort of mini method fingerprints themselves).
- FieldFilter: Class/static field usage, by defining class, field name, field type.
- LiteralFilter: Literal instruction of type long/double
- Logical OR filter: Match one of any filters. Useful when instructions differ between multiple app targets.

Projects can define their own custom instruction filters, such as ResourceMappingPatch with it's own kind of LiteralFilter that matches resource literal values (no more mucking about with using a ResourcePatches to first set a resource value a fingerprint then uses).
&nbsp;

#### Variable space allowed between instruction filters
By default, all filters allow any amount of space between them.  But if filters are always immediately after each other, or there is a rough estimate of the maximum number of indexes until the next instruction, then a maximum distance can be set.  An example is using an opcode filter of `MOVE_RESULT` or `MOVE_RESULT_OBJECT` after a method call, where the max instruction spacing is always 0.
&nbsp;

### Breaking changes
Fuzzy pattern match is now obsolete, as it's functionality is now part of the filtering itself.  Variable spacing is allowed between instruction filters, and non important instructions are now ignored by simply not defining instruction filters for them.

&nbsp;

### Example fingerprint before and after this change
Before:
```

val bottomBarContainer = resourceMappings["id", "bottom_bar_container"]

internal val shortsBottomBarContainerFingerprint = fingerprint {
    accessFlags(AccessFlags.PUBLIC, AccessFlags.FINAL)
    returns("V")
    parameters("Landroid/view/View;", "Landroid/os/Bundle;")
    strings("r_pfvc")
    literal { bottomBarContainer }
}

shortsBottomBarContainerFingerprint.method.apply {
    // Search for indexes after the fact, after the fingerprint already resolved.

    // First instruction of interest.
    val resourceIndex = indexOfFirstLiteralInstruction(bottomBarContainer)
    
    // Second instruction of interest.
    val index = indexOfFirstInstructionOrThrow(resourceIndex) {
        getReference<MethodReference>()?.name == "getHeight"
    }
    
    // Third instruction of interest.
    val heightRegister = getInstruction<OneRegisterInstruction>(index + 1).registerA
    
[...]

PR #385, opened 2025-11-18
feat: Modernize APIs

oSumAtrIX:
# About

This PR gets rid of the fingerprint system and replaces it with a simpler API to find methods and classes. It offers a DSL API for that purpose. On top it simplifies internal code handling mutable classes/proxies & adds new apis for creating patches with their name obtained from the property to reduce duplicating the name in the property and name parameter.

What to expect:

image

May supersede https://github.com/ReVanced/revanced-patcher/pull/329 or be merged with it.

# Todo

- Modernize more places of the patcher, e.g, the patcher apis.
- Finish code
- Implement tests
- Implement caching (how?)
- IMPORTANT: Right now, the matcher api is built in a way where it will be instantiated everytime the lambdas are called which is inefficient. The matchers need to be built once.

### DIFF
--- a/patcher/src/commonMain/kotlin/app/revanced/patcher/Matching.kt
+++ b/patcher/src/commonMain/kotlin/app/revanced/patcher/Matching.kt
@@ -1121,7 +1121,7 @@ operator fun String.invoke(compare: String.(String) -> Boolean = String::equals)
 operator fun Opcode.invoke(): IndexedMatcherPredicate<Instruction> =
     { _, _, _ -> opcode == this@invoke }
 
-val primitiveTypes = setOf("V", "Z", "B", "S", "C", "I", "J", "F", "D")
+internal val primitiveTypes = setOf("V", "Z", "B", "S", "C", "I", "J", "F", "D")
 
 // Returns a comparer function based on the format of a type descriptor string.
 private fun String.typeComparer(): String.(String) -> Boolean = when {


oSumAtrIX:
> You wrongly criticize using custom instruction filters as "punching a hole in the API" (instruction filter is the API!)

It is a correct criticism. Using a custom block goes beyond the defined API boundaries. Supplying the block itself is part of the API, however, the logic inside it bypasses the API entirely. That effectively creates a hole in the abstraction. Furthermore, relying on this escape mechanism for certain scenarios results in inconsistent API usage, which should not be necessary if the API were fully designed.

> silly non-existent use cases such as checking if a string has an odd number of letters,

I have repeatedly explained that this is an example demonstrating the API’s inability to express all matching cases. You have continued to miss that point. The example is valid regardless of how trivial it seems to you. It represents a legitimate scenario that forces a fallback to the custom block.

> then you casual explain your PR here basically requires abusing itself to just get things done.

You ignored the core of what I stated: determining string indices **is not** the task of method matching. Any API can be misused beyond its intended purpose, including both yours and mine. I showed that string matching can still be forced to work using the API, which is outside of its design. Your response misrepresents this and implies my API only functions through misuse for its intended purpose, which is incorrect.

> I have a simple solution that is proven to work. All versions of YT from 20.01 to 20.46 patch without using custom filters or the existing fingerprint kludge block. It's a simple yet flexible expansion of opcode matching with the introduction of variable spacing between the relevant parts of the method instructions.

Possibly, but your approach introduces numerous regressions that you have omitted from your comparison. They are documented here: [https://github.com/ReVanced/revanced-patcher/pull/385#issuecomment-3551916424](https://github.com/ReVanced/revanced-patcher/pull/385#issuecomment-3551916424). My solution addresses all listed issues, while yours does not. Your API appears “proven” only because you have had over a year to refactor patches around its limitations. My proposal has existed for only a short time, and I have already shown that it is capable of handling arbitrary matching scenarios. Suggesting otherwise is misleading.

> We can merge what I have, then you can do v23 patcher that is your vision of blocks of code logic everywhere that provides no benefits (none!) over what I already have working right now.

Your current API is inconsistent and opinionated. Declaring that it provides “no” benefits compared to mine directly contradicts the detailed list of API improvements I already provided.

> I'm still waiting

As stated multiple times, around five by now, the information you requested is already present in the other PR comments.

> Nobody will be able to post anything because instruction filter can handle it all

You rely on the API’s escape hatch,  the custom block,  to achieve matching. That is not the API itself. The existence of this unrestricted raw lambda block proves the API does **not** handle every case. Nothing prevents placing **all** logic inside the custom block, skipping your API entirely.

>METHODS THAT CANNOT BE FINGERPRINTED WITH INSTRUCTION FILTERS WITHOUT USING CUSTOM CLASSES AND WITHOUT CUSTOM FINGERPRINT BLOCKS

That said, if I provide an example APK in which your API fails to meet the following requirements, no additional workarounds should be considered acceptable, like we've seen being raised multiple times across the review. If you agree, send the commit of your PR that is the latest, and I'll send the conditions.

oSumAtrIX:
Now just this needs implementation lol

image

oSumAtrIX:
A couple of example Instruction. extensions:

image

image

oSumAtrIX:
So right now, I have no "builder" behind all these APIs. This leads to the issue of not being able to cache evaluations. Previously I solved this by provoding an instance to the outer APIs where you could cache your data. I have implemented an API "remember" which can be used to cache expressions. The .match APIs for example use this already so their builder scopes dont get reevaluated. I just noticed that this same API is useful in other scopes too. E.g. you have a very expensive computation you could otherwise cache:

image

The example shows that the expression inside the remember lambda can technically be written as a val outside of firstMethod, but say if you need the scope of a parent, (e.g. in your expensive computation you use the method scopes name) then remember is useful for caching here.

Caching is what has to be manually be done by the API user, so it is not ideal, but i think there is no way around it since the API cant know what to evaluate and what not (unless you write a small helper, like the match apis)

oSumAtrIX:
image

Here are examples of the match api

oSumAtrIX:
I figured another great usecase of the API during developmenet is that you can inject any code such as a println for debugging purposes or even place breakpoints.

oSumAtrIX:
image

Discovered a bug, this does not handle labels. 

image

This should fix it.

oSumAtrIX:
image

and with this we have a handful of easy apis. This should be feature complete apart from two missing things:

labels/offsets & reference2 of a DualReferenceInstruction, though i cant imagine a way to use offset, since you can match for `is<OffsetInstruction>` & reference2 is difficult to scale since it requires duplicating all apis to X2 apis, e.g. reference2() and method2()

PR #411, opened 2026-03-19
add regression tests for methodsByString cache population

mostafaNazari702:


patcher/src/jvmTest/kotlin/app/revanced/patcher/MatchingTest.kt
```diff
@@ -232,4 +237,114 @@ class MatchingTest : PatcherTestBase() {
...
+            assertNotNull(
+                classDefs.methodsByString["This is a test."],
+                "Expected methodsByString to contain 'This is a test.' after initialization",
+            )
+            assertNull(
+                classDefs.methodsByString["nonexistent string"],
+                "Expected methodsByString to not contain a string that doesn't exist",
+            )
+
+            // verify string-based method lookup works using the context function.
```
oSumAtrIX: I dont think this test is correct. Nothing says that calling firstImmutableMethodOrNull actually makes use of the cache

oSumAtrIX:
The commit mentions

> Fixes: methodsByString not being set during initialzation.

But where does it fix it?

The only API that I think can cause to data inconsistency is the addAll api (iirc). I have left a comment in the class explaining why.

PR #445, opened 2026-06-17
fix: sanitize fake png resources before compile

rahmibabapro:
## Summary
- Adds a pre-AAPT resource sanitizer for decoded APK resources.
- Scans `res/**/*.png`, validates the PNG magic bytes, detects JPEG/WebP/AVIF/HEIF content, and renames fake `.png` files to a supported Android drawable extension when safe.
- Keeps valid `.png` and `.9.png` resources unchanged; invalid `.9.png` files fail with an actionable `PatchException` instead of reaching AAPT2 as a generic PNG signature error.
- Adds JVM tests for fake PNG detection, valid PNG preservation, nine-patch preservation/rejection, and a minimal fake `bs6.png` fixture that is sanitized to `bs6.webp` before compile.

## Root cause
Some APKs can contain decoded drawable files with a `.png` extension while the actual bytes are another image format, such as WebP. AAPT2 validates by extension and fails with `failed to read PNG signature`. The resource is already in the patcher's decoded resource directory, so the robust fix belongs in `revanced-patcher` before `AaptInvoker.invoke`, not in Manager UI handling.

## Behavior
- PNG magic bytes: preserve.
- JPEG/WebP/AVIF: rename to `.jpg`, `.webp`, or `.avif`, preserving the resource name such as `@drawable/bs6`.
- HEIF/unknown or invalid `.9.png`: fail with a clear message naming the file, detected format, and suggested action.

## Testing
- `git diff --cached --check` passed before commit.
- `./gradlew :patcher:jvmTest` could not complete in this environment because the available GitHub token lacks `read:packages` for `app.revanced:apktool-lib:2.10.1.1` and `app.revanced:multidexlib2:3.0.3.r3` from GitHub Packages (401 Unauthorized).

## Notes
- I could not update the separate TikTok `Disable telemetry` fingerprint because `revanced/revanced-patches` currently returns GitHub DMCA 403 from this environment.

### DIFF
--- a/patcher/src/commonMain/kotlin/app/revanced/patcher/patch/ResourceFileSanitizer.kt
+++ b/patcher/src/commonMain/kotlin/app/revanced/patcher/patch/ResourceFileSanitizer.kt
@@ -1,0 +1,141 @@
+package app.revanced.patcher.patch
+
+import java.io.File
+import java.nio.file.Files
+import java.util.Locale
+import java.util.logging.Logger
+import java.util.stream.Collectors
+
+internal object ResourceFileSanitizer {
+    private val pngMagic = byteArrayOf(0x89.toByte(), 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A)
+    private val jpegMagic = byteArrayOf(0xFF.toByte(), 0xD8.toByte(), 0xFF.toByte())
+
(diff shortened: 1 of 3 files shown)


oSumAtrIX:
Decoding is handled by apktool-lib. Why is this not implemented there?

